PR #14, opened 2024-12-18
Feature/bflat

valentinbreiz:


Liquip.Sdk/Sdk.targets
```diff
@@ -0,0 +1,40 @@
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+	<PropertyGroup>
+		<PatcherPath>$(MSBuildProjectDirectory)\..\Liquip.Patcher\bin\Debug\net8.0\Liquip.Patcher.exe</PatcherPath>
```
kaleb-itm, line 4: you want to use $(MSBuildThisFile) then it can work from a nuget package

Guillermo-Santos:
Add `dotnet tool install ilc -g` before build

kaleb-itm:
we should be able to add it as a dependency as use the executable from there

PR #30, opened 2025-01-29
fix: fix cicd

zarlo:


zarlo:
i ran the commands on linux and they are returned a 0 status

PR #35, opened 2025-02-06
start of base of ASM building

zarlo:


valentinbreiz:
hello @zarlo, should not we focus on x64 since NativeAOT barely supports x32? I think we should focus on x64 and ARM

ascpixi:
Are we _sure_ we need the X# builder API surface? While it would be nice, I think it'd be a lot more code to maintain, especially given the fact that x86-64 has a lot of ISA extensions itself. If we're limiting inline assembly to the minimum, wouldn't X# be extremely underutilized in this scenario?

I believe that we could get by if we only did what is currently known as `LiteralAssemblerCode` in IL2CPU. That is, when defining an inline assembly method, we would do the following:

```cs
[InlineAssembly]
public static void ReadCr3() {
  InlineAssembly.Literal("""
    mov rax, cr3
    ret
  """);
}
```

zarlo:
@valentinbreiz ill keep working on the build tasks i am not (build asm and link the objects in to one) as we will need it either way

zarlo:
im going to wait for #43 so this can be easily tested

PR #38, opened 2025-02-12
Feature/formatting pre hook

kumja1:


zarlo:
@kumja1 we should enforce it at PR too or it will be hard to enforce as a user might not have git hooks run

kumja1:
@zarlo  try and see if the git hooks work for you when you pull this. I set the hook folder to .githooks instead of .git/hooks in the repo's config

zarlo:
ill have time in a few hours or tommrow

PR #42, opened 2025-02-16
upd: improve analyzer and codefix

kumja1:


kumja1:
@zarlo  is it fine if we check for formating only on pull request  not pushes to the main  branch

Guillermo-Santos:
> @zarlo is it fine if we check for formating only on pull request not pushes to the main branch

I think there is a setting you can enabled so all commits of a branch have to come from a pull request, so that is not a bad idea.

zarlo:
> > @zarlo is it fine if we check for formating only on pull request not pushes to the main branch
> 
> I think there is a setting you can enabled so all commits of a branch have to come from a pull request, so that is not a bad idea.

yep we (or well @valentinbreiz ) can make it so you cant push to main

PR #43, opened 2025-02-27
Feature/ilc

valentinbreiz:


zarlo:
This does not need to be done in this PR, but in #35 i have a nuget package that has a linker for windows and im using LdPath to get the path, im happy to change it to CppLinker

Guillermo-Santos:
Using the dotnet bat file worked for me, but if there are other ways then it may be useful.

Guillermo-Santos:
should we change to a nuget patckage reference instead of calling the target directly?

kumja1:
@valentinbreiz  This is ready to merge

PR #47, opened 2025-03-11
Feature/ilc tests

kumja1:


Guillermo-Santos:
print the current working directory on main.cpp, it may be different

PR #49, opened 2025-03-22
Refactor/project namespace

kumja1:


kumja1:
@valentinbreiz

tests/Cosmos.NativeWrapper/Cosmos.NativeWrapper.csproj
```diff
@@ -1,5 +1,15 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net8.0</TargetFramework>
```
zarlo: this should be managed by Directory.Build.props so it can be removed

kumja1: This should be resolved @zarlo

PR #50, opened 2025-03-29
Cosmos.Kernel.Tests

valentinbreiz:


Guillermo-Santos:
As the output of Cosmos.ilc.build changed, we need to also change the test to reflect those changes

valentinbreiz:
> As the output of Cosmos.ilc.build changed, we need to also change the test to reflect those changes

Yes but we want to have the POC compiling before

tests/Cosmos.Kernel.Tests/Kernel.csproj
```diff
@@ -0,0 +1,33 @@
...
+        <TargetFramework>net8.0</TargetFramework>
+        <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+
+        <NoStdLib>true</NoStdLib>
+        <NoConfig>true</NoConfig>
+        <RuntimeMetadataVersion>v4.0.30319</RuntimeMetadataVersion>
+        <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
+        <GenerateTargetFrameworkAttribute>false</GenerateTargetFrameworkAttribute>
+
+        <IlcSystemModule>Kernel</IlcSystemModule>
```
Guillermo-Santos: `IlcSystemModule` may need to be Cosmos.Kernel (the assembly's name)

Guillermo-Santos: A change should be also made on [ILC.Build.targets](https://github.com/valentinbreiz/nativeaot-patcher/edit/feature/Cosmos.Kernel.Tests/src/Cosmos.ilc.Build/ILC.Build.targets) on the `WriteILCRsp` target, to set the `--systemmodule` parameter if `IlcSystemModule` is specified.

src/Cosmos.ilc.Build/ILC.Build.targets
```diff
@@ -89,6 +90,7 @@
       <IlcArg Include="--generateunmanagedentrypoints:System.Private.CoreLib" />
       <IlcArg Include="--export-dynamic-symbol:DotNetRuntimeDebugHeader" />
       <IlcArg Include="--export-unmanaged-entrypoints" />
+      <IlcArg Condition="$(IlcSystemModule) != ''" Include="--systemmodule:$(OutputPath)$(IlcSystemModule).dll" />
```
Guillermo-Santos: This should be `<IlcArg Condition="$(IlcSystemModule) != ''" Include="--systemmodule:$(IlcSystemModule)" />` the dll is/should be on the references already

kumja1: Doesn't work

Guillermo-Santos: The corelib should probably be on another project

Guillermo-Santos: That or it should be Comos.Kernel

kumja1: > That or it should be Comos.Kernel

I don't think so, the output dll in the bin is Kernel.dll

Guillermo-Santos: Oh

Guillermo-Santos: I will check in a moment

Guillermo-Santos:
now this is the error
image

kumja1:
now the error is this
image

PR #53, opened 2025-04-02
feat: Demonstrate Limine requests in the kernel PoC + build rules improvements

ascpixi:
Functionally speaking, this PR demonstrates the ability to use Limine requests directly in C# code using pre-initialized statics.

### Summary
* Updated the SDK version to `8.0.100` in `global.json`.
* Added the kernel PoC to the solution.
* Adjusted build rules to locate `ilc` manually (as opposed to with `PublishAot`).
* Cleaned up the build rules for the PoC.


ascpixi:
cc @valentinbreiz

src/Cosmos.ilc.Build/ILC.Build.targets
```diff
@@ -47,24 +81,40 @@
...
 
-    <Exec Command="ld.lld -T src/linker.ld -m elf_x86_64 -nostdlib -static -z max-page-size=0x1000 $(MSBuildProjectDirectory)/bin/$(Configuration)/net8.0/$(RuntimeIdentifier)/native/$(AssemblyName).a -o $(MSBuildProjectDirectory)/bin/$(Configuration)/net8.0/$(RuntimeIdentifier)/publish/$(AssemblyName).elf" />
+    <ItemGroup>
+      <LinkerArgs Include="-T src/linker.ld" />
+      <LinkerArgs Include="-m elf_x86_64" />
+      <LinkerArgs Include="-nostdlib" />
+      <LinkerArgs Include="-static" />
+      <LinkerArgs Include="-z max-page-size=0x1000" />
+      <LinkerArgs Include="%(ManagedBinary.OutputFile)" />
+      <LinkerArgs Include="-o $(MSBuildProjectDirectory)/bin/$(Configuration)/net8.0/$(RuntimeIdentifier)/publish/$(AssemblyName).elf" />
```
Guillermo-Santos: you can change this to `<LinkerArgs Include="-o $(OutputPath)/$(AssemblyName).elf" />`

So it would not depend on .NET version.

valentinbreiz: ping me when this is solved

ascpixi: This does actually change the actual path of the output ELF file, but I feel like the updated path is actually more appropriate than the one we had before. This doesn't really matter either way, to be honest, as this is a proof-of-concept.

ascpixi: @valentinbreiz Solved :shipit:

Issue #57, opened 2025-04-03
Fix iso-tests action (missing ilc)


valentinbreiz:
https://github.com/valentinbreiz/nativeaot-patcher/actions/runs/14250541817/job/39941990087#step:9:5471

Guillermo-Santos:
I have seen that sometimes the `DependsOnTargets` property is not respected and run without it.

[https://github.com/valentinbreiz/nativeaot-patcher/blob/7a43505ea36128b0449a9120bd46899d3def76a1/src/Cosmos.ilc.Build/ILC.Build.targets#L91](https://github.com/valentinbreiz/nativeaot-patcher/blob/7a43505ea36128b0449a9120bd46899d3def76a1/src/Cosmos.ilc.Build/ILC.Build.targets#L91)

Guillermo-Santos:
I see the test also install the ilc workload. That means that we could make something like this
```xml
<IlcToolsPath Condition="$(IlcHostPackagePath) != ''">$(IlcHostPackagePath)/tools/</IlcToolsPath>
```
And change the exec to
```xml
    <Exec
      Command="&quot;$(IlcToolsPath)ilc&quot; @&quot;%(ManagedBinary.RspFile)&quot;"
      WorkingDirectory="$(FullIntermediateOutputPath)"
    />
```

Removed an `/` on the call, since we are already adding it on the `IlcToolsPath` definition.

So, if the target did not found it, the workload should do it.

Issue #55, opened 2025-04-03
Create a dedicated task for ISO creation


valentinbreiz:
Now everything is in https://github.com/valentinbreiz/nativeaot-patcher/blob/main/src/Cosmos.ilc.Build/ILC.Build.targets

Resources:
https://github.com/valentinbreiz/nativeaot-patcher/blob/main/src/Cosmos.Asm.Build/Tasks/LdTask.cs

PR #61, opened 2025-04-04
Fix iso-tests action

Guillermo-Santos:
#57


Guillermo-Santos:
Wonder, can we make a task that do the same thing as the workload? that way at least ilc compilation could be self contained (on our package)

valentinbreiz:
> Wonder, can we make a task that do the same thing as the workload? that way at least ilc compilation could be self contained (on our package)

Yes  https://github.com/valentinbreiz/nativeaot-patcher/issues/55

Issue #33, opened 2025-02-03
Make folders for Cosmos Gen3


zarlo:
To make it less work to split up later we should put the current src and test folder in a folder named Liquip, make a folder called Cosmos and Asm (or XSharp) with stubs in both folder

like this
```
.
├── Cosmos
│   ├── src
│   └── test
├── XSharp
│   ├── src
│   └── test
├── Liquip
│   ├── src
│   └── test
├── Directory.Build.props
├── Directory.Packages.props
└── README.md
```

PR #64, opened 2025-04-07
Implement X# building

valentinbreiz:
Definition of done:
- [x] Logic for X# 64bit
- [x] Logic for X# ARM
- [x] dotnet 9
- [x] project structure rework according to #33 
- [x] Implement X# 64bit
- [x] Implement assembly plugs in C# (Cosmos.Patcher) using XSharp task
- [x] Emit XSharp assembly to 64bit assembly
- [x] Build assembly using yasm
- [x] Link obj to kernel


9xbt:
Why?

valentinbreiz:
> Why?

Because X# can implement ARM8 and x64 and that is was we need for #6

valentinbreiz:
> > > Why?
> > 
> > 
> > Because X# can implement ARM8 and x64 and that is was we need for #6
> 
> We are only going to use architecture-specific assembly methods, I don't really see a point in doing this?

X# is an HLA (high level assembly language) for x86 and ARM assembly. We want to do assembly plugs in X# like before 
eg. https://github.com/CosmosOS/IL2CPU/blob/master/source/Cosmos.Core.DebugStub/CmdSend.xs#L159

ascpixi:
Excuse my ignorance, but I still don't understand the "high-level assembly" part of X#. I always understood X# to be an alternative assembly notation (common across architectures), with some features like variables and whatnot (which I assume are more so aliases for registers, as we'd want to control the registers we're actually using). This would be more akin to an assembly flavor.

What would be high-level in X# compared to regular assembly?

valentinbreiz:
It aims to simplify operating system development by incorporating C-like language syntax to assembly language. Initially, X# was used for debugging services in Cosmos. The X# compiler is a command-line interface program that parses code lines into tokens, compares them with patterns, and translates matched patterns to x86 or ARM assembly.

valentinbreiz:
> > https://github.com/CosmosOS/IL2CPU/blob/master/source/Cosmos.Core.DebugStub/CmdSend.xs#L159
> 
> There is no point in doing this all in assembly, since if we implement `outb` and `inb` functions in Cosmos we can then implement SendText in C# for example.

I agree with that

PR #70, opened 2025-04-09
Fix hardcoded ASM linkage and replace with something more modular.

AzureianGH:
This stops the ASM hardcoding of the test OS.

(My git is acting up so if something is weird just leave a comment)


AzureianGH:
Lots of commits because github desktop didn't want to register any of them, but just check file changes.

PR #78, opened 2025-04-09
Replace as much ASM as possible with managed C# Code.

AzureianGH:


Cosmos/src/Asm/io.asm
```diff
@@ -0,0 +1,67 @@
...
+global _native_io_read_word
+global _native_io_read_dword
+global _native_io_read_qword
+
+section .text
+
+; void out_byte(ushort port, byte value)
+_native_io_write_byte:
+    push rdx
+    push rax
```
ascpixi: Under System V, both RDX and RAX are scratch registers, and thus do not need to be preserved.

This applies to all of the functions declared here.

Cosmos/src/Internal.cs
```diff
@@ -0,0 +1,77 @@
...
+            [RuntimeImport("*", "_native_io_read_word")]
+            public static extern ushort Read16(ushort Port);
+
+            [MethodImpl(MethodImplOptions.InternalCall)]
+            [RuntimeImport("*", "_native_io_read_dword")]
+            public static extern uint Read32(ushort Port);
+        }
+    }
+
+    public unsafe static class Serial
```
ascpixi: Instead of marking the entire class as `unsafe`, we should only mark the methods that make use of unsafe code as such instead.

ascpixi (changes requested):
LGTM, a few minor nit-picks. Given that this relates to the (temporary) kernel proof-of-concept, this isn't that serious, but the changes requested should be low-hanging fruit.

Cosmos/src/Asm/io.asm
AzureianGH: I got used to the x86 COSMOS calling from my elf stuff so I just put those there without thinking

Cosmos/src/Internal.cs
AzureianGH: To note, the original Serial class (I didn't make it) was marked as `unsafe`.

Cosmos/src/Asm/io.asm
AzureianGH: Fixed

Cosmos/src/Internal.cs
AzureianGH: Fixed

ascpixi (approved):
LGTM, there's some inconsistencies in regard to our general code style but given that this is the PoC that doesn't really matter :shipit:

valentinbreiz:
Cool changes! Does it work?

PR #79, opened 2025-04-10
Refactor Cosmos.Asm.Build

Guillermo-Santos:
It now starts to follow what is shown on the [wiki](https://github.com/valentinbreiz/nativeaot-patcher/wiki).


Guillermo-Santos:
Next thing to make, move output of Cosmos.Asm.Build to `obj` instead of `bin`.

Maybe to `$(IntermediateOutputPath)/Cosmos/Asm`

PR #90, opened 2025-05-02
Feat: Build Kernel using Cosmos nuget packages

Guillermo-Santos:
This break test and CI, you would first need to create a folder with the next route from the root of the repo `./artifacts/package/debug` since there is where packages of the repo are posted (on debug mode), then you would need to build Cosmos.Patcher and all Cosmos.*.Build projects to generate the nuget packages, then you can build the sample kernel


Guillermo-Santos:
Should I Fix CI on this PR too?

zarlo:
fix it if you can

Guillermo-Santos:
> fix it if you can

Then i will look at it tomorrow.

Guillermo-Santos:
I think this is ready

PR #91, opened 2025-05-14
Feacture/bcl

valentinbreiz:


src/Cosmos.Ilc.Build/build/Cosmos.Ilc.Build.targets
```diff
@@ -51,6 +51,7 @@
       <NativeOutputPath>$([System.IO.Path]::GetFullPath('$(OutputPath)/cosmos/native/'))</NativeOutputPath>
       <IlcIntermediateOutputPath>$(FullIntermediateOutputPath)/cosmos/native/</IlcIntermediateOutputPath>
       <FinalAssembly>$(NativeOutputPath)$(AssemblyName)</FinalAssembly>
+      <IlcSystemModule Condition="$(IlcSystemModule) == ''">System.Private.CoreLib</IlcSystemModule>
```
valentinbreiz, line 45: I don't get why IlcSystemModule is System.Private.CoreLib if we are using Test.CoreLib

Guillermo-Santos: That is just the default value, on the kernel example we override it with `Test.CoreLib`

valentinbreiz: Ok nevermind it is overrided in Kernel.csproj

Guillermo-Santos:
Whot? Iso builds failed? but it worked on fine on the devcontainer and all changes were committed

PR #96, opened 2025-05-27
Fix hard-coded values from wrong project.

AzureianGH:
I accidentally pushed the testing one instead of the final (My bad)


selkij:
the number of warnings is huge

PR #98, opened 2025-07-02
Feature/bcl rework

valentinbreiz:


kumja1:
> Will Comos.Kernel.Runtime be the main package the user needs to reference?

I think it will Cosmos.Kernel.System

Guillermo-Santos:
> I think it will Cosmos.Kernel.System
Perfect, the reason ISO Build Test is failing is because some projects referencing Cosmos.Kernel.Runtime, i'll see if i can remove the "NoCoreLib".

src/Cosmos.Patcher/MonoCecilExtension.cs
```diff
@@ -1815,57 +1815,47 @@ public static void OptimizeInstructions(this MethodDefinition method)
...
-        {
-            Console.WriteLine($"[GetArgument] Processing fallback: {indexOrStr} ({indexOrStr.GetType().Name})");
-
-            CustomAttributeArgument? argument = indexOrStr switch
-            {
-                int index => index > attribute.ConstructorArguments.Count - 1 || index < 0
-                    ? null
-                    : attribute.ConstructorArguments[index],
-                string name => attribute.Properties.FirstOrDefault(arg => string.Equals(arg.Name, name, StringComparison.InvariantCultureIgnoreCase)).Argument,
-                _ => null
```
Guillermo-Santos, line 1835: One of the problems is with changes on this part, it always returning null

Guillermo-Santos:
@valentinbreiz everything is fixed now, you can merge.

PR #97, opened 2025-06-27
Add Page and Heap Allocators

zarlo:
this for the most part is a port from Cosmos Gen2

this only has the allocators for now

i will see if i can hook it in the .nets runtime

Issue #92, opened 2025-05-14
Add Base Class Library project


valentinbreiz:
Test.Corelib (or [System.Private.Corelib](https://github.com/valentinbreiz/nativeaot-patcher/issues/104)) should be (two solutions) : 

- [x] ~~Cross compiled manually for windows + linux support with dlls pushed in the repo directly~~
- [ ] Build BCL from sources (Create a git submodule of dotnet runtime)
- [x] Integrate BCL binary into Builder (+ CI)
- [x] ~~Add the dll in nuget package~~

valentinbreiz:
Also we still use a dll directly, we should build our own coreLib from sources.

valentinbreiz:
> On [#98](https://github.com/valentinbreiz/nativeaot-patcher/pull/98) we got a successful build without Test.CoreLib, is this still needed?

It may be needed since nativeAOT System.Private.CoreLib is not directly made for baremetal. Some parts from BCL may changes.

valentinbreiz:
I close this for now, can be reopened if build BCL from sources is needed

PR #107, opened 2025-08-20
Implement basic Rhp allocation using MemoryOp

valentinbreiz:
## Summary
- integrate MemoryOp allocator into runtime
- implement basic Rhp helpers (RhpNewFast, RhpNewArray, RhpAssignRef) for string allocation
- reference Cosmos.Kernel.Core project in Cosmos.Kernel.Runtime
- drop Core's runtime reference to avoid circular build

## References
- https://github.com/dotnet/dotnet/blob/main/src/runtime/src/coreclr/nativeaot/Runtime/portable.cpp

### DIFF
--- a/src/Cosmos.Kernel.Runtime/Stdllib.cs
+++ b/src/Cosmos.Kernel.Runtime/Stdllib.cs
@@ -223,7 +223,7 @@ namespace Internal.Runtime.CompilerHelpers
         {
             return RhpNewArray(pEEType, length);
         }
-        
+
         /*
                 [RuntimeExport("RhTypeCast_CheckCastAny")]
                 static unsafe object RhTypeCast_CheckCastAny(object obj, int typeHandle) { throw null; }
--- a/src/Cosmos.Kernel.Runtime/Stdllib.cs
+++ b/src/Cosmos.Kernel.Runtime/Stdllib.cs
@@ -2,6 +2,7 @@
 
 using System;
 using System.Runtime;
+using Cosmos.Kernel.Core.Memory;
 
 namespace System
 {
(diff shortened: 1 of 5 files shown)


src/Cosmos.Kernel.Runtime/Stdllib.cs
```diff
@@ -122,6 +123,17 @@ public RuntimeImportAttribute(string dllName, string entry)
...
+        internal unsafe struct MethodTable
+        {
+            internal ushort _usComponentSize;
+            private ushort _usFlags;
+            internal uint _uBaseSize;
+            internal MethodTable* _relatedType;
+            private ushort _usNumVtableSlots;
+            private ushort _usNumInterfaces;
+            private uint _uHashCode;
+        }
```
Guillermo-Santos, line 136: Since we are going to have a dotnet as a github submodule, we may want to replace this and add the real MethodTable struct directly from dotnet source using the csproj.

valentinbreiz: You're right, it will be done in https://github.com/valentinbreiz/nativeaot-patcher/issues/92

zarlo: my PR will have this https://gitlab.com/liquip/nativeaot-patcher/-/commit/6810738623fbda3677f5c0cabcae1a267c1215dd?page=2#d4570af40b74aa51d1f1dc2e5f912330dd727bf7_11_0 is what we will want to do, there is a file that has 1 method that we will need to add that not included from the repo

valentinbreiz: Perfect

valentinbreiz:
Status: Fault on VMWare

valentinbreiz:
Waiting for #97

PR #110, opened 2025-08-20
Port Cosmos Heap

zarlo:
this is a port of the cosmos heap and page Allocators

### DIFF
--- a/src/Cosmos.Kernel.Runtime/Memory/PageAllocator.cs
+++ b/src/Cosmos.Kernel.Runtime/Memory/PageAllocator.cs
@@ -65,8 +65,8 @@ public static unsafe class PageAllocator
     /// </exception>
     public static void InitializeHeap(byte* aStartPtr, ulong aSize)
     {
-        Serial.WriteString("UART started.\n");
-        
+        Serial.WriteString("[PageAllocator] UART started.\n");
+
         if ((uint)aStartPtr % PageSize != 0)
         {
             throw new Exception("RAM start must be page aligned.");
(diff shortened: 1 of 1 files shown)


zarlo:
nooooo

valentinbreiz:
Some Rhp implementations seems needed

zarlo:
yes for the PR its just the Allocators not the runtime integration

Guillermo-Santos:
> Some Rhp implementations seems needed

Some of those are implemented, others should have been removed by the runtime features, seems like something isn't running? i can check in 2 hours to see what happened.

valentinbreiz:
Blocked by #111

src/Cosmos.Kernel.Runtime/Stdllib.cs
```diff
@@ -95,9 +95,15 @@ private static unsafe void RhpCheckedAssignRef(void** location, void* value)
...
-
+        [RuntimeExport("RhpCheckedXchg")]
+        private static void* InterlockedExchange(void** location1, void* value)
+        {
+            void* original = *location1;
+            *location1 = value;
+            return original;
+        }
+      
+        /*
```
Guillermo-Santos: Remove this line, it was added for some reason lol

Guillermo-Santos: line 106, there is a multiline comment is opened, but never closed.

src/Cosmos.Kernel.Runtime/MetaTable.cs
```diff
@@ -0,0 +1,41 @@
...
+    internal static unsafe int RhGetModuleFileName(IntPtr moduleHandle, out byte* moduleName) {
+        moduleName = (byte*) 0x00;
+        return 0;
+    }
+
+    [RuntimeExport("RhpCheckedXchg")]
+    internal static object InterlockedExchange([NotNullIfNotNull(nameof(value))] ref object? location1, object? value)
+    {
+        return value;
+    }
```
Guillermo-Santos: you may want to transfer the implementation that is here:
https://github.com/valentinbreiz/nativeaot-patcher/blob/0fb4dacd0463ba1c14ca93f5900f1a3d9cbf1ee9/src/Cosmos.Kernel.Runtime/Stdllib.cs#L99

One of the two should be removed or build will fail.

9xbt:
Doesn't the old cosmos heap have issues? I remember the GC eating my kernel.

zarlo:
> Doesn't the old cosmos heap have issues? I remember the GC eating my kernel.

yes, big time but the Heap implementation is fine

zarlo:
i dont know when i will have time to work on this so its up for grabs

src/Cosmos.Kernel.Runtime/MetaTable.cs
```diff
@@ -0,0 +1,41 @@
...
+public class MetaTable
+{
+    // look int what this does
+    [RuntimeExport("RhpGetModuleSection")]
+    private static IntPtr RhGetModuleSection(ref object module, object section,
+        out int length)
+    {
+        length = 0;
+        return IntPtr.Zero;
+    }
```
Guillermo-Santos: This is already implemented in main.

https://github.com/valentinbreiz/nativeaot-patcher/blob/main/src/Cosmos.Kernel.Runtime/ModuleHelpers.cs#L11

src/Cosmos.Kernel.Runtime/Memory.cs
```diff
@@ -0,0 +1,77 @@
+using System.Runtime;
+using Cosmos.Kernel.Core.Memory;
```
Guillermo-Santos, line 2: Add 
```csharp
using Internal.Runtime;
```
We may even add it as a global using.

PR #122, opened 2025-08-23
Cosmos.Patcher fixes

valentinbreiz:
- [x] scan only if matched plug
- [x] don't add all targets, only plugreference
- [x] add logger

### DIFF
--- a/src/Cosmos.Patcher/PlugPatcher.cs
+++ b/src/Cosmos.Patcher/PlugPatcher.cs
@@ -2,6 +2,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Xml.Linq;
 using Cosmos.Build.API.Attributes;
+using Cosmos.Patcher.Logging;
 using Mono.Cecil;
 using Mono.Cecil.Cil;
 using Mono.Cecil.Rocks;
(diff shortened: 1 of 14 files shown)


src/Cosmos.Patcher/PlugPatcher.cs
```diff
@@ -57,6 +101,18 @@ public void PatchAssembly(AssemblyDefinition targetAssembly, params AssemblyDefi
...
 
+    private static string? GetPlugTargetName(TypeDefinition plugType)
+    {
+        CustomAttribute? plugAttr = plugType.CustomAttributes
+            .FirstOrDefault(attr => attr.AttributeType.FullName == typeof(PlugAttribute).FullName);
+        if (plugAttr == null)
+            return null;
+
+        return plugAttr.ConstructorArguments.Count == 1 && plugAttr.Properties.Count == 0
+            ? plugAttr.GetArgument<string>()
```
kumja1, line 116: @valentinbreiz `plugAttr.ConstructorArguments.Count == 1 && plugAttr.Properties.Count == 0
            ? plugAttr.GetArgument<string>()`. This is unecessary as GetArgument will check for positional if the named argument isn't found or provided.

Issue #121, opened 2025-08-23
RFC: Project Refrance Layout


zarlo:
to prevent circular reference we should have a strict reference order

```mermaid
flowchart LR;
	UsersKernel-->Cosmos.Kernel;
	UsersKernel-->Cosmos.Kernel.System;
	UsersKernel-->Cosmos.Kernel.HAL;
	UsersKernel-->Cosmos.Kernel.Core;
    Cosmos.Kernel-->Cosmos.Kernel.Boot.*;
	Cosmos.Kernel-->Cosmos.Kernel.SubSystem;
	Cosmos.Kernel.SubSystem-->Cosmos.Kernel.System;
	Cosmos.Kernel.SubSystem-->Cosmos.Kernel.HAL;
    Cosmos.Kernel-->Cosmos.Kernel.System;
    Cosmos.Kernel.System-->Cosmos.Kernel.System.ARM/x86;
	Cosmos.Kernel.System.Plug-->Cosmos.Kernel.System;
    Cosmos.Kernel.System-->Cosmos.Kernel.HAL;
	Cosmos.Kernel.HAL-->Cosmos.Kernel.HAL.ARM/x86;
	Cosmos.Kernel.HAL.Plug-->Cosmos.Kernel.HAL;
	Cosmos.Kernel.HAL-->Cosmos.Kernel.Core;
    Cosmos.Kernel.Core-->Cosmos.Kernel.Core.ARM/x86;
	Cosmos.Kernel.Core.Plug-->Cosmos.Kernel.Core;
```

Cosmos.Kernel.SubSystem is things like Cosmos.Kernel.System.Graphics

Issue #106, opened 2025-08-20
Disable unsafe code by default


valentinbreiz:
Disable GCC generation by default, let user enable it from project configuration

PR #125, opened 2025-08-24
Add Cosmos.Kernel aggregator and enforce boot reference rules

valentinbreiz:
Fixes https://github.com/valentinbreiz/nativeaot-patcher/issues/121

## Summary
- create Cosmos.Kernel project aggregating core, HAL, system, graphics, runtime, plugs and boot
- enforce boot reference layering with Directory.Build.targets
- document kernel project layout and update example to reference new aggregator


Guillermo-Santos:
if we are going to follow the graph in #121 to the T, then `Cosmos.Kernel.System.Graphics` and `Cosmos.Kernel.HAL` have it's dependencies reversed, so `Cosmos.Kernel.HAL` should define the interfaces/abstract classes and `Cosmos.Kernel.System.Graphics` implement it.

Guillermo-Santos:
This is something extra, but we can also make `Cosmos.Kernel` hold all the references to `Cosmos.Build.*` projects needed, there is a way to make then reference only but that they don't take part on `Cosmos.Kernel` building. That way we could make the `Cosmos.Sdk` be like Aspire one, just some glue between the packages and that enforce the need to reference the main package (`Cosmos.Kernel` on this case).

And it would finally be compatible with global package management (it isn't right now).

valentinbreiz:
I think this PR could be used to fix https://github.com/valentinbreiz/nativeaot-patcher/issues/106 too. Every C or ASM (except Rh stubs) sources should be kept in early boot phases before we jump to C#.

valentinbreiz:
> @valentinbreiz 
> So, renaming `kmain` to `bootstrap` in the C and linker files is the only thing missing on this PR, or is there any other missing feature?
 I made some code reviews too

PR #130, opened 2025-08-24
clean up project files

zarlo:


valentinbreiz:
Maybe not related to this pull request but why we still need to use netstandard instead of net9.0 everywhere?

valentinbreiz (changes requested):
Patcher and Scanner test may need project configuration updates

zarlo:
some of the project need to be netstandard

PR #157, opened 2025-09-17
Implement some module's Rh functions.

Guillermo-Santos:
This PR do some things:

1. Add's dotnet/runtime as a git submodule.
2. Reference the MethodTable struct from it.
3. Add new types to interact with modules
4. Implement some of the Rh functions used to interact with modules.


Guillermo-Santos:
Don't know if it is needed right now, or if it will ever be needed, so do not merge right now.

valentinbreiz:
> Don't know if it is needed right now, or if it will ever be needed, so do not merge right now.

We should start working on runtime integration as a git submodule so yes if it works we should merge this instead of using System.Private.CoreLib

PR #168, opened 2025-10-03
Extra gcc flags for windows

Guillermo-Santos:


Guillermo-Santos:
Let's see if this works (can't test my selft lol)

PR #169, opened 2025-10-04
Add BIOS font support

AzureianGH:
Since PSF isn't working right now, I added BIOS font support until I make abstract fonts.

### DIFF
--- a/src/Cosmos.Kernel.System.Graphics/Canvas.cs
+++ b/src/Cosmos.Kernel.System.Graphics/Canvas.cs
@@ -83,16 +83,16 @@ public unsafe class Canvas
 
     public static void DrawChar(char c, int x, int y, uint color)
     {
-        PCScreenFont.PutChar(c, x, y, color, Color.Transparent);
+        BIOSFont.PutChar(c, x, y, color, Color.Transparent);
     }
 
     public static void DrawString(string text, int x, int y, uint color)
     {
-        PCScreenFont.PutString(text, x, y, color, Color.Transparent);
+        BIOSFont.PutString(text, x, y, color, Color.Transparent);
     }
 
     public static unsafe void DrawString(char* text, int x, int y, uint color)
     {
-        PCScreenFont.PutString(text, x, y, color, Color.Transparent);
+        BIOSFont.PutString(text, x, y, color, Color.Transparent);
     }
 }
--- a/src/Cosmos.Kernel.System.Graphics/Fonts/BIOSFont.cs
+++ b/src/Cosmos.Kernel.System.Graphics/Fonts/BIOSFont.cs
@@ -1,0 +1,106 @@
+using Cosmos.Kernel.Core.Memory;
+using Cosmos.Kernel.System.Conversion;
+
+namespace Cosmos.Kernel.System.Graphics.Fonts;
+
+public static unsafe class BIOSFont
+{
+    public static class Default
+    {
+        public static int Size = 4096; // Bytes array
+
+        public static char* GetUnmanagedFontData()
+        {
+            // 8x16 BIOS font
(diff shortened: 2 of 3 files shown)


src/Cosmos.Kernel.Native.x64/CPU/RhpStackProbe.asm
```diff
@@ -0,0 +1,28 @@
...
+;;   r11 - preserved
+;; Notes:
+;;   Probes at least one page below rsp.
+global RhpStackProbe
+section .text
+
+
+PROBE_STEP equ 0x1000
+
+RhpStackProbe:
```
ascpixi: This should be a no-op in kernel-mode!

src/Cosmos.Kernel.System.Graphics/Canvas.cs
```diff
@@ -83,16 +85,69 @@ public static void ClearScreen(uint color)
...
+        DefaultFont.PutString(text, x, y, color, Color.Transparent);
     }
 
     public static unsafe void DrawString(char* text, int x, int y, uint color)
     {
-        PCScreenFont.PutString(text, x, y, color, Color.Transparent);
+        DefaultFont.PutString(text, x, y, color, Color.Transparent);
+    }
+
+    public static unsafe void DrawQuadraticBezier(uint color, int x0, int y0, int x1, int y1, int x2, int y2)
```
ascpixi: Are we sure we want to do this in kernel mode...? Why would we need Bezier curve drawing in the kernel?

src/Cosmos.Kernel.System.Graphics/Fonts/BIOSFont.cs
```diff
@@ -0,0 +1,109 @@
...
...Y2NjY2NjcwPwAAAAAAAAAAAAAAAAA/MDc2NjY2NjY2NjY2NjY29wD/AAAAAAAAAAAAAAAAAP8A9zY2NjY2NjY2NjY2NjY3MDc2NjY2NjY2NgAAAAAA/wD/AAAAAAAAAAA2NjY2NvcA9zY2NjY2NjY2GBgYGBj/AP8AAAAAAAAAADY2NjY2Njb/AAAAAAAAAAAAAAAAAP8A/xgYGBgYGBgYAAAAAAAAAP82NjY2NjY2NjY2NjY2NjY/AAAAAAAAAAAYGBgYGB8YHwAAAAAAAAAAAAAAAAAfGB8YGBgYGBgYGAAAAAAAAAA/NjY2NjY2NjY2NjY2NjY2/zY2NjY2NjY2GBgYGBj/GP8YGBgYGBgYGBgYGBgYGBj4AAAAAAAAAAAAAAAAAAAAHxgYGBgYGBgY/////////////////////wAAAAAAAAD////////////w8PDw8PDw8PDw8PDw8PDwDw8PDw8PDw8PDw8PDw8PD/////////8AAAAAAAAAAAAAAAAAAHbc2NjY3HYAAAAAAAB4zMzM2MzGxsbMAAAAAAAA/sbGwMDAwMDAwAAAAAAAAAAA/mxsbGxsbGwAAAAAAAAA/sZgMBgwYMb+AAAAAAAAAAAAftjY2NjYcAAAAAAAAAAAZmZmZmZ8YGDAAAAAAAAAAHbcGBgYGBgYAAAAAAAAAH4YPGZmZjwYfgAAAAAAAAA4bMbG/sbGbDgAAAAAAAA4bMbGxmxsbGzuAAAAAAAAHjAYDD5mZmZmPAAAAAAAAAAAAH7b29t+AAAAAAAAAAAAAwZ+29vzfmDAAAAAAAAAHDBgYHxgYGAwHAAAAAAAAAB8xsbGxsbGxsYAAAAAAAAAAP4AAP4AAP4AAAAAAAAAAAAYGH4YGAAA/wAAAAAAAAAwGAwGDBgwAH4AAAAAAAAADBgwYDAYDAB+AAAAAAAADhsbGBgYGBgYGBgYGBgYGBgYGBgYGNjY2HAAAAAAAAAAABgYAH4AGBgAAAAAAAAAAAAAdtwAdtwAAAAAAAAAOGxsOAAAAAAAAAAAAAAAAAAAAAAAABgYAAAAAAAAAAAAAAAAAAAAGAAAAAAAAAAADwwMDAwM7GxsPBwAAAAAANhsbGxsbAAAAAAAAAAAAABw2DBgyPgAAAAAAAAAAAAAAAAAfHx8fHx8fAAAAAAAAAAAAAAAAAAAAAAAAAAAAA==";
```
ascpixi: I don't think we should include fonts out-of-the-box. This is fine if temporary, but ideally we'd remove this in the future.

src/Cosmos.Kernel.System.Graphics/Fonts/BIOSFont.cs
```diff
@@ -0,0 +1,109 @@
...
+
+    public override int CharWidth => 8;
+
+    public override int CharHeight => 16;
+
+    private static void EnsureInitialized()
+    {
+        if (!_Initialized)
+        {
+            byte* fontData = Base64.Decode(Default.GetUnmanagedFontData(), (uint)Default.Size);
```
ascpixi: We _should_ be able to use a `ReadOnlySpan` to store the bytes directly, without the need for Base64-decoding. For example:

https://github.com/ascpixi/smolsharp/blob/54a4aeb6e8f6cdd64d049862dee611301082e71b/src/SmolSharp.Ocean/Shaders/OceanShader.cs#L20-L56

ascpixi (changes requested):
Nicely done! We do probably need some changes on this, though.

src/Cosmos.Kernel.System.Graphics/Canvas.cs
AzureianGH: It was a sanity check to make sure floats weren't being wacky, I forgot to remove it

src/Cosmos.Kernel.Native.x64/CPU/RhpStackProbe.asm
AzureianGH: I didn't know that raghhhhh I was just getting an error

src/Cosmos.Kernel.System.Graphics/Fonts/BIOSFont.cs
AzureianGH: It's so we can have some form of display text output that isn't serial

src/Cosmos.Kernel.System.Graphics/Fonts/BIOSFont.cs
AzureianGH: Fonts were made awhile before we could, and base64 works for now so.. 🤷‍♂️ I never bothered changing it since it worked, but if you think readonlyspan works, be my guest!

PR #171, opened 2025-10-12
Fix/string builder

valentinbreiz:


src/Cosmos.Kernel.Core/Memory/PageAllocator.cs
```diff
@@ -232,129 +219,240 @@ public static void InitializeHeap(byte* aStartPtr, ulong aSize)
...
             for (ulong i = 0; i < Limine.MemoryMap.Response->EntryCount; i++)
             {
                 LimineMemmapEntry* entry = Limine.MemoryMap.Response->Entries[i];
                 if (entry->Type == LimineMemmapType.Usable)
                 {
-                    // Convert physical address to virtual address
-                    ulong virtualStart = (ulong)entry->Base + 0xFFFF800000000000UL;
+                    ulong physStart = (ulong)entry->Base;
+                    ulong physEnd = physStart + entry->Length;
+                    ulong virtualStart = physStart + 0xFFFF800000000000UL;
```
Guillermo-Santos, line 269: Why only the largest usable region instead of all?

src/Cosmos.Kernel.Core/Memory/PageAllocator.cs
```diff
@@ -232,129 +219,240 @@ public static void InitializeHeap(byte* aStartPtr, ulong aSize)
...
+                    // Check if this region overlaps with framebuffer
+                    bool overlapsFramebuffer = false;
+                    if (framebufferStart > 0 && framebufferEnd > 0)
+                    {
+                        if (physStart < framebufferEnd && physEnd > framebufferStart)
+                        {
+                            overlapsFramebuffer = true;
+                            Serial.WriteString(" [OVERLAPS FRAMEBUFFER]");
+                        }
+                    }
```
Guillermo-Santos, line 301: Are these checks really needed? Usable regions wouldn't overlap with any of this two, as they are on their respective regions (as far as I know)

src/Cosmos.Kernel.Core/Memory/PageAllocator.cs
valentinbreiz: Because it needs more update in allocator to be able jumping between available regions

src/Cosmos.Kernel.Core/Memory/PageAllocator.cs
valentinbreiz: It's to not overwrite kernel with heap

Guillermo-Santos:
Is this finished or the allocator will be updated more?

PR #173, opened 2025-10-18
Port over Cosmos Pci

zarlo:


src/Cosmos.Kernel.Core/IO/Serial.cs
```diff
@@ -115,4 +115,58 @@ public static void WriteHex(uint number)
...
+            {
+                case null:
+                    WriteString(NULL);
+                    break;
+                case string s:
+                    WriteString(s);
+                    break;
+                case char c:
+                    WriteString(c.ToString());
+                    break;
```
Guillermo-Santos, line 135: I think it may be best to add a `WriteChar` method than do this

valentinbreiz: ToString is indeed buggy atm

Guillermo-Santos: ToString load some type metafata for lost types, i saw it break because of an infinite loop in Lock.TryEnterSlow, that is called somewhere in the Encoding classes

valentinbreiz:
Status: Seems to crash at LoadIdt(&ptr);

valentinbreiz (approved):

src/Cosmos.Kernel/Kernel.cs
```diff
@@ -54,6 +60,14 @@ public static unsafe void Initialize()
             Serial.WriteString("Architecture: Unknown.\n");
         }
 
+        InterruptManager.Initialize();
+        PciManager.Setup();
+        int a = 0;
```
zarlo: we should remove this

PR #174, opened 2025-10-19
Add Some Rh functions implementations

Guillermo-Santos:
This adds some of the required Rh functions for boxing + some other useful ones


valentinbreiz:
> > looks to be broken on arm
> 
> Arm build is broken from some time ago, can fix it, actually

i think it would be better to fix ARM in another pull request

Edit: Nevermind it's already fixed ahah

valentinbreiz:
I'm having troubles with string lists, should I continue here or this pr is ready to merge?

zarlo (approved):

PR #177, opened 2025-10-31
Fix ARM kernel

valentinbreiz:


examples/DevKernel/DevKernel.csproj
```diff
@@ -15,6 +15,10 @@
     <InvariantGlobalization>true</InvariantGlobalization>
     <SelfContained>true</SelfContained>
     <PublishTrimmed>true</PublishTrimmed>
+
+    <!-- Disable stack trace metadata to avoid StackTraceMetadata initialization crash -->
+    <IlcGenerateStackTraceData>false</IlcGenerateStackTraceData>
+    <StackTraceSupport>false</StackTraceSupport>
```
Guillermo-Santos, line 21: This flags dont do anything right now, you can test if removing them works, or add support to them.

Guillermo-Santos: Forget it, tested and and StackTraceSupport = false to avoid the initializer

Guillermo-Santos (approved):
Nice

PR #195, opened 2025-11-18
Feature/typecasting

kumja1:


Guillermo-Santos:
@kumja1 add a new runtime export called `RhpStackProbe`, it can be noop
```csharp
[RuntimeExport("RhpStackProbe")]
private static void RhpStackProbe() 
{
}
```

examples/DevKernel/DevKernel.csproj
```diff
@@ -3,11 +3,11 @@
   <Sdk Name="Cosmos.Sdk" Version="3.0.0" />
 
   <PropertyGroup>
-    <OutputType>Exe</OutputType>
+    <OutputType>Library</OutputType>
```
Guillermo-Santos: Revert this change, then it will work

PR #197, opened 2025-12-02
Add project template

valentinbreiz:
- Add cosmos-kernel template for `dotnet new` with x64/arm64 support
- Fix ILC path resolution for standalone nuget.org package usage
- Add PackageDownload for ILCompiler to ensure correct .NET 10 ILC
- Exclude System.Private.CoreLib from patching (copy unchanged)
- Use absolute paths in Cosmos.Build.Ilc.targets
- Rename ResolveIlcPath to ResolveIlcToolsPath to avoid target conflict
- Update package versions for nuget.org release

### DIFF
--- a/.devcontainer/postCreateCommand.sh
+++ b/.devcontainer/postCreateCommand.sh
@@ -18,113 +18,81 @@ fi
 echo "Using define: $ARCH_DEFINE"
 echo "Using runtime: $RUNTIME_ID"
 
-# CRITICAL: Clean everything to avoid stale file references
-echo "Cleaning all build artifacts and caches..."
-
-# Clear all NuGet locals cache first
-dotnet nuget locals all --clear
-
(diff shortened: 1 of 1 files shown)


Guillermo-Santos:
> * Exclude System.Private.CoreLib from patching (copy unchanged)

Are you sure about this one? shouldn't we be able to patch System.Private.CoreLib?

valentinbreiz:
> > * Exclude System.Private.CoreLib from patching (copy unchanged)
> 
> Are you sure about this one? shouldn't we be able to patch System.Private.CoreLib?

I'm currently investigating, it's opus hallucination xD

valentinbreiz:
But it says this because roslyn does not include plug if they are not needed...

valentinbreiz:
> But most of the current plug are on System.Private.CoreLib xd

Dev kernel seems to build

kumja1 (approved):

PR #207, opened 2025-12-13
Add E1000E driver + network manager service

valentinbreiz:


src/Cosmos.Kernel.HAL.X64/Pci/PciManager.cs
```diff
@@ -140,8 +141,9 @@ public static bool Exists(PciDevice pciDevice) =>
     /// <returns></returns>
     public static PciDevice? GetDevice(uint bus, uint slot, uint function)
     {
-        foreach (PciDevice xDevice in Devices)
```
zarlo, line 143: speed wise this is the same as the compiler does this for you

valentinbreiz: You mean build time or execution time?

PR #215, opened 2025-12-30
Implement multithreading

valentinbreiz:


src/Cosmos.Kernel.Core/Memory/Heap/Heap.cs
```diff
@@ -17,23 +18,30 @@ public static unsafe class Heap
     /// <returns>New pointer with specified size while maintaining old data.</returns>
     public static byte* Realloc(byte* aPtr, uint newSize)
     {
+        InternalCpu.DisableInterrupts();
```
zarlo: we should make a class so we can wrap this in a using block to auto re-enable on an error

src/Cosmos.Kernel.Core/Scheduler/Thread.cs
```diff
@@ -0,0 +1,85 @@
...
+{
+    // ===== Identity =====
+    public uint Id { get; set; }
+    public uint CpuId { get; set; }
+
+    // ===== State =====
+    public ThreadState State { get; set; }
+    public ThreadFlags Flags { get; set; }
+
+    // ===== Context (architecture-specific values) =====
```
zarlo, line 16: context item should not be publicly editable

src/Cosmos.Kernel.Core/Scheduler/ThreadContext.X64.cs
```diff
@@ -0,0 +1,124 @@
+using System.Runtime.InteropServices;
```
zarlo, line 1: for this we should mirror what MS does and have a ThreadContext.X64.cs and ThreadContext.ARM64.cs

src/Cosmos.Kernel.Plugs/System/Threading/ThreadPlug.cs
```diff
@@ -0,0 +1,169 @@
...
+    private static void ThreadEntryPoint()
+    {
+        // Now try to get thread info
+        var cpuState = SchedulerManager.GetCpuState(0);
+        var currentThread = cpuState.CurrentThread;
+
+        if (currentThread == null)
+        {
+            Serial.WriteString("[ThreadPlug] No current thread!\n");
+            while (true) { }
```
zarlo: do we want this to panic the system as this will just burn cpu time and the core could still be in a broken state

src/Cosmos.Kernel.Plugs/System/Threading/ThreadPlug.cs
```diff
@@ -0,0 +1,169 @@
...
+        Serial.WriteNumber(threadId);
+        Serial.WriteString("\n");
+
+        // Get and invoke the delegate
+        if (_threadDelegates.TryGetValue(threadId, out var start))
+        {
+            _threadDelegates.Remove(threadId);
+
+            Serial.WriteString("[ThreadPlug] Invoking delegate\n");
+            start.Invoke();
```
zarlo: this can throw we should wrap it so we can clean up the thread and kill it with a none 0 code

valentinbreiz: fixed in ac5af2ab92198b5ab71b2d1c63fcbdd0dbb96b1d with exit code in serial.

src/Cosmos.Kernel.Plugs/System/Threading/ThreadPlug.cs
valentinbreiz: can a throw be enough?

src/Cosmos.Kernel.Core/Scheduler/ThreadContext.X64.cs
valentinbreiz: Since for HAL we use `Cosmos.Kernel.HAL.X64` and `Cosmos.Kernel.HAL.ARM64` naming,
Should we use `Cosmos.Kernel.Core.X64 `and `Cosmos.Kernel.Core.ARM64` instead of using separate files in the same project? 
Or should we keep the seperation in the same project like you are proposing with `ClassName.X64.cs` and `ClassName.ARM64.cs`?
Or maybe we should even merge every Core source in HAL and have something like Services -> HAL -> Native

kumja1: I really like the MS approach. It seems strange having a separate project for what is essentially duplicates of the same file but for different architecture's

valentinbreiz: Okay, then HAL would need a rewrite for consistency?

src/Cosmos.Kernel.Plugs/System/Threading/ThreadPlug.cs
Guillermo-Santos: maybe an argument null exception

src/Cosmos.Kernel.Core/Scheduler/ThreadContext.X64.cs
Guillermo-Santos: if there were to be a rewrite then it would only be in PortIO and the CPU Ops, since they can be static and be rewrriten as partial classes, but the other things...they would be in ther `x64` or `arm64` namespace.

valentinbreiz: Done in [9c0c041](https://github.com/valentinbreiz/nativeaot-patcher/pull/215/commits/9c0c04175d2c6da4e011a70cd8e77ffc4eb3e72c)

src/Cosmos.Kernel.Plugs/System/Threading/ThreadPlug.cs
valentinbreiz: Kernel panic done in [68c4872](https://github.com/valentinbreiz/nativeaot-patcher/pull/215/commits/68c4872ee61dbe657b23ee42d95e2e019402eef6)

src/Cosmos.Kernel.Core/Memory/Heap/Heap.cs
valentinbreiz: a class for `InternalCpu.DisableInterrupts();`? I don't understand
Or you are takling about Dispose?

src/Cosmos.Kernel.Core/Scheduler/Thread.cs
valentinbreiz: Done in eb6bb93d482aff0cf5f20a327655bb440e4aee0e

src/Cosmos.Kernel.Core/Memory/Heap/Heap.cs
Guillermo-Santos: > a class for `InternalCpu.DisableInterrupts();`? I don't understand Or you are takling about Dispose?

I think he is saying about having a type that implements IDisposable that in the dispose method reenable interrups, so we can have a scope for the code that requires the interrups to be disabled. Like this:

```csharp
using (var scope = InternalCpu.DisableInterruptsScope()) // Disable Interrups and return an IDisposable
{
    // Some Code Here
}
// Interrups are reenable automatically by the dispose.
```

Guillermo-Santos: if we go by this then this type must have a static integer field that is increased by one each time it's ctor is called and decreased by one each time the Dispose method is called and interrups are reenabled only when this count is 0, this is so that if a method that start a scope is called inside one that also start an scope, interrups aren't reenable until we dispose in the caller.

zarlo: > > a class for `InternalCpu.DisableInterrupts();`? I don't understand Or you are takling about Dispose?
> 
> I think he is saying about having a type that implements IDisposable that in the dispose method reenable interrups, so we can have a scope for the code that requires the interrups to be disabled. Like this:
> 
> ```cs
> using (var scope = InternalCpu.DisableInterruptsScope()) // Disable Interrups and return an IDisposable
> {
>     // Some Code Here
> }
> // Interrups are reenable automatically by the dispose.
> ```

This

PR #228, opened 2026-01-16
Implement Cosmos Graphic Subsystem

valentinbreiz:


src/Cosmos.Kernel.Core/Memory/ManagedMemoryBlock.cs
```diff
@@ -0,0 +1,267 @@
+﻿using System;
+
+namespace Cosmos.Kernel.Core.Memory;
+
+/// <summary>
+/// ManagedMemoryBlock class. Used to read and write a managed memory block.
+/// </summary>
+public unsafe class ManagedMemoryBlock
```
zarlo, line 8: i think we should use Span<T> over this

src/Cosmos.Kernel.HAL.Interfaces/Devices/IGraphicDevice.cs
```diff
@@ -0,0 +1,17 @@
+// This code is licensed under MIT license (see LICENSE for details)
+
+namespace Cosmos.Kernel.HAL.Interfaces.Devices;
+
+/// <summary>
+/// Interface for timer devices.
+/// </summary>
+public interface IGraphicDevice
```
zarlo, line 10: there should be a method to take an array or a Span<T> for mass update

src/Cosmos.Kernel.Core/Memory/ManagedMemoryBlock.cs
Guillermo-Santos: Span<> or Memory<>

valentinbreiz: Done in 04c890af48d146353415eec88f3b7b8135894028

src/Cosmos.Kernel.HAL.Interfaces/Devices/IGraphicDevice.cs
valentinbreiz: Done in 721628e3212fc7d6ef49e64306514bac162dd2cb

PR #232, opened 2026-01-21
Implement interface dispatch resolution

valentinbreiz:
See memory test kernel

### DIFF
--- a/src/Cosmos.Build.Ilc/build/Cosmos.Build.Ilc.targets
+++ b/src/Cosmos.Build.Ilc/build/Cosmos.Build.Ilc.targets
@@ -122,18 +122,12 @@
       <IlcArg Include="-O"/>
       <!-- Generate map file for debugging symbol resolution -->
       <IlcArg Include="--map:$(IlcIntermediateOutputPath)$(AssemblyName).map.xml"/>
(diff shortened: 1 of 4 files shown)


src/Cosmos.Build.Ilc/build/Cosmos.Build.Ilc.targets
```diff
@@ -122,9 +122,6 @@
       <IlcArg Include="-O"/>
       <!-- Generate map file for debugging symbol resolution -->
       <IlcArg Include="--map:$(IlcIntermediateOutputPath)$(AssemblyName).map.xml"/>
-      <!-- Architecture-specific instruction sets -->
-      <IlcArg Include="--instruction-set:base" Condition="'$(CosmosArch)' == 'x64'"/>
-      <IlcArg Include="--instruction-set:base" Condition="'$(CosmosArch)' == 'arm64'"/>
```
Guillermo-Santos, line 127: what problem was this causing?

valentinbreiz: Nothing i just cleaned we don't use CosmosArch anymore

PR #235, opened 2026-01-28
Implement Garbage Collector

valentinbreiz:
Implement garbage collector with mark and sweep phases, using [Cosmos Memory Manager SMT (Size Map Table)](https://cosmosos.github.io/articles/Kernel/MemoryManagement.html) blocks 

Thanks @kevingosse for his wonderful [Writing a .NET Garbage Collector in C#](https://minidump.net/writing-a-net-gc-in-c-part-6/) serie.

Documentation available at https://valentinbreiz.github.io/nativeaot-patcher/articles/garbage-collector.html

Todo:
- [x] Implement ARM
- [x] Fix 0x100000 hardcoded value
- [x] Fix collect causing page fault on new allocations
- [x] Add garbage collection tests

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -8,7 +8,7 @@
     },
     "features": {
         "ghcr.io/devcontainers/features/dotnet:2": {},
-        "ghcr.io/devcontainers/features/powershell:1": {}
+        "ghcr.io/devcontainers/features/powershell:2": {}
     },
 	"hostRequirements": {
 		"cpus": 4,


src/Cosmos.Kernel.Core/Memory/GarbageCollector.cs
```diff
@@ -0,0 +1,747 @@
...
+    /// <summary>
+    /// Number of pointers in this series.
+    /// </summary>
+    public uint NumPtrs;
+
+    /// <summary>
+    /// Bytes to skip after these pointers.
+    /// </summary>
+    public uint Skip;
+}
```
Guillermo-Santos: weren't this two contained in a single value? i think this struct is two times the correct size

kevingosse: They're stored on half-pointers. So using two `uint` fields is valid on 64 bit and invalid on 32 bit. Depends on what runtimes you're targeting.

zarlo: 64 bit only ARM and x86

valentinbreiz:
> What about that error in arm64?

what error? the ISO build tests?

PR #250, opened 2026-02-08
start on async Serial

zarlo:


.devcontainer/postCreateCommand.sh
```diff
@@ -133,7 +133,7 @@ rm -rf ~/.nuget/packages/cosmos.* 2>/dev/null || true
 
 # Restore main solution
 echo "Restoring main solution..."
-dotnet restore ./nativeaot-patcher.slnx
```
Guillermo-Santos: I think this solutions isn't updated, or is it?

src/Cosmos.Kernel.Core/Internal/Runtime/CompilerHelpers/LibraryInitializer.cs
```diff
@@ -26,6 +26,8 @@ public static void InitializeLibrary()
             // Initialize managed modules
             Serial.WriteString("[KERNEL]   - Initializing managed modules...\n");
             ManagedModule.InitializeModules();
+
+            Serial.StartThread();
```
Guillermo-Santos: Move this to the library initializer in Cosmos.Kernel, that's the place where the scheduler setup is finished.

PR #262, opened 2026-02-23
Add RingMaster analyzer

valentinbreiz:
image

### DIFF
--- a/src/Cosmos.Build.Analyzer.Patcher/PatcherAnalyzer.cs
+++ b/src/Cosmos.Build.Analyzer.Patcher/PatcherAnalyzer.cs
@@ -262,6 +262,11 @@ namespace Cosmos.Build.Analyzer.Patcher
 
             DebugLog($"Checking plug class name for target {name} (simple: {simpleClassName})");
             AnalyzePlugClass(plugClass, simpleClassName, context);
+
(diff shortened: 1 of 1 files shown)


src/Cosmos.Build.Analyzer.Patcher/LayerAnalyzer.cs
```diff
@@ -0,0 +1,161 @@
...
+            context.RegisterCompilationAction(AnalyzeCompilation);
+        }
+
+        /// <summary>
+        /// Returns true for assemblies named *.Plugs — plug projects bridge all layers
+        /// and are exempt from layer checks by convention.
+        /// </summary>
+        private static bool IsPlugAssembly(string assemblyName)
+            => assemblyName.EndsWith(".Plugs", System.StringComparison.Ordinal)
+            || assemblyName.Equals("Plugs", System.StringComparison.Ordinal);
```
Guillermo-Santos, line 53: Why don't we use [CompilerVisibleProperty](https://github.com/dotnet/roslyn/blob/393a9937dca86abd44f82b109204a29f1b3ae556/docs/features/source-generators.cookbook.md?plain=1#L541) instead? and have a property like `IsCosmosPlug` set to true or false that we can read from the analyzer.

PR #266, opened 2026-03-02
Add minor changes

Demiomad:
These are things that are visible by a regular user after this pull request gets merged:
- The ability to change the graphical console's font using `KernelConsole.Font`, this was done because I thought that the default font was too big for the graphical console.
- `VersionMinor` inside "Cosmos.Kernel.Kernel" is 38 rather than 0 (This is done in order to sync with the version string, most people just use the version string)

Other things that aren't visible by a regular user:
- `CustomFont.psf` was renamed to `DefaultFont.psf`, this was done because it makes more sense.
- Line 181 on ThreadPlug.cs is now:
```cs
var exitThread = exitCpuState.CurrentThread;
```
rather than:
```cs
var exitThread = exitCpuState?.CurrentThread;
```
because of "Conditional access qualifier expression is never null according to nullable reference types' annotations"
I actually didn't want to commit this, but I forgot to add the question mark back. It makes no difference anyway lmao

### DIFF
--- a/src/Cosmos.Kernel/Kernel.cs
+++ b/src/Cosmos.Kernel/Kernel.cs
@@ -12,7 +12,7 @@ public class Kernel
     // CosmosOS version - keep in sync with kmain.h
     public const int VersionMajor = 3;
     public const int VersionMinor = 0;
-    public const int VersionPatch = 38;
+    public const int VersionPatch = 0;
     public const string VersionString = "3.0.38";
     public const string Codename = "gen3";
     public const string VersionBanner = $"CosmosOS v{VersionString} ({Codename}) - Managed runtime active\n";


src/Cosmos.Kernel/Kernel.cs
```diff
@@ -12,7 +12,7 @@ public class Kernel
     // CosmosOS version - keep in sync with kmain.h
     public const int VersionMajor = 3;
     public const int VersionMinor = 0;
-    public const int VersionPatch = 0;
+    public const int VersionPatch = 38;
```
Guillermo-Santos: 38 is the current minor version in the nuget packages, this should be reverted

Demiomad: Ah, I didn't know, thank you

PolishBoiYT:
DONT MERGE YET I MISTOOK MAJOR FOR PATCH IN THE "Revert *MAJOR*_VERSION" COMMIT

PolishBoiYT:
i cant change the commit message

Demiomad:
wtf happened here 😭

PR #267, opened 2026-03-02
Add minor changes

Demiomad:
These are things that are visible by a regular user after this pull request gets merged:
- The ability to change the graphical console's font using `KernelConsole.Font`, this was done because I thought that the default font was too big for the graphical console.

Other things that aren't visible by a regular user:
- `CustomFont.psf` was renamed to `DefaultFont.psf`, this was done because it makes more sense.
- Line 181 on ThreadPlug.cs is now:
```cs
var exitThread = exitCpuState.CurrentThread;
```
rather than:
```cs
var exitThread = exitCpuState?.CurrentThread;
```
because of "Conditional access qualifier expression is never null according to nullable reference types' annotations"
I actually didn't want to commit this, but I forgot to add the question mark back. It makes no difference anyway lmao


src/Cosmos.Kernel.Plugs/System/ConsolePlug.cs
```diff
@@ -266,4 +266,4 @@ private static ConsoleKey MapToConsoleKey(ConsoleKeyEx key)
             _ => ConsoleKey.NoName
         };
     }
-}
```
Guillermo-Santos, line 269: a newline at the end of each file is currently a code format prerequisite, add it again and everything will be fine

PR #269, opened 2026-03-03
Remove invalid options from "cosmos new" command

Demiomad:
**The reason why this was done is because of this error:**
```ansi
  Creating Cosmos Kernel Project
  --------------------------------------------------
  Name: tt
  Architecture: x64
  Graphics: Yes
  --------------------------------------------------

  [unning: dotnet new cosmos-kernel -n tt -o /home/plboiyt/cosmos/gen3/nativeaot-patcher/.test/tt --TargetArch x64 --EnableGraphics true

  Failed to create project: Error: Invalid option(s):
--TargetArch
   '--TargetArch' is not a valid option
x64
   'x64' is not a valid option
--EnableGraphics
   '--EnableGraphics' is not a valid option
true
   'true' is not a valid option

For more information, run:
   dotnet new cosmos-kernel -h

For details on the exit code, refer to https://aka.ms/templating-exit-codes#127


  If the template is not found, install it with:
    dotnet new install Cosmos.Build.Templates
```

The `dotnet new` template didn't support those options, the simplest fix was to remove the options from the `cosmos new` command until they are supported.

### DIFF
--- a/src/Cosmos.Tools/Commands/NewCommand.cs
+++ b/src/Cosmos.Tools/Commands/NewCommand.cs
@@ -15,6 +15,7 @@ public class NewSettings : CommandSettings
     [Description("Output directory (defaults to current directory)")]
     public string? Output { get; set; }
 
+    /*
     [CommandOption("-a|--arch")]
     [Description("Target architecture (x64, arm64)")]
     [DefaultValue("x64")]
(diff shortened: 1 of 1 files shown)


valentinbreiz:
> I checked the source code for the VSCode extension, it doesn't seem to use the `cosmos new` command.
> https://github.com/valentinbreiz/CosmosVsCodeExtension/blob/main/src/commands/newProject.ts#L133
> 
> Unless I'm missing something

Oh indeed my bad. Then it can be removed

PR #288, opened 2026-03-15
Add install scripts + Release CI  (bundled win-mac-linux binaries)

valentinbreiz:


setup/Cosmos.iss
```diff
@@ -0,0 +1,250 @@
...
+
+; NuGet packages
+Source: "bundle\packages\*.nupkg"; DestDir: "{app}\Packages"; Flags: ignoreversion
+
+; Cross-compiler toolchains
+Source: "bundle\tools\windows\x86_64-elf-tools\*"; DestDir: "{app}\Tools\x86_64-elf-tools"; Flags: ignoreversion recursesubdirs createallsubdirs
+Source: "bundle\tools\windows\aarch64-elf-tools\*"; DestDir: "{app}\Tools\aarch64-elf-tools"; Flags: ignoreversion recursesubdirs createallsubdirs
+
+; Build tools
+Source: "bundle\tools\windows\yasm\*"; DestDir: "{app}\Tools\yasm"; Flags: ignoreversion recursesubdirs createallsubdirs
```
zarlo, line 55: we can put this all in a nuget package (make a new package for it like Cosmos.BuildTools.Bin)

valentinbreiz: You mean three packages for all host architecture? Or win + linux + mac in a big multi rid package?
Maybe three packages but definitly not a big package it would be really huge

zarlo: 1 for each so Cosmos.BuildTools.Bin.[Platform], we should likely omit linux because of compatibility issues with libc versions and kernel versions but looking at how nuget pull in packages i think it will pull them ll down no matter what but it might be able to be done with a custom workload

Guillermo-Santos: Or it may be downloaded by the Sdk but not added as a package reference

valentinbreiz: > Or it may be downloaded by the Sdk but not added as a package reference

If we force a tool version per cosmos versions yes because I don't want us to support tool versions issues if tools have new versions.

valentinbreiz: > 1 for each so Cosmos.BuildTools.Bin.[Platform], we should likely omit linux because of compatibility issues with libc versions and kernel versions but looking at how nuget pull in packages i think it will pull them ll down no matter what but it might be able to be done with a custom workload
> 

Don't nuget will cause issues with a lot of pre compilated binaries in nuget packages? Because of security issues

valentinbreiz: > > Or it may be downloaded by the Sdk but not added as a package reference
> 
> If we force a tool version per cosmos versions yes because I don't want us to support tool versions issues if tools have new versions.

Also I want a all in one release for Win users to have everything packaged in installer

Guillermo-Santos: or maybe the win installer should already have all the binaries? other platforms can install the required tools 🤔

valentinbreiz: > or maybe the win installer should already have all the binaries? other platforms can install the required tools 🤔

That's what i started with CI but I should drop "manual" CI tool download and download in CI from cosmos install command anyway, with fixed versions in cosmos install

PR #328, opened 2026-04-17
Work on Serial driver

zarlo:
Serial is a real hot path so I'm going to try and make it not block

- [x] async driver (none blocking)
- [ ] overload generator for Write(...)
- [ ] find things that i can update to the async driver (needs heap and threading)


zarlo:
will wait on #340

PR #340, opened 2026-05-03
Unified kernel logging system

### LINKED ISSUE #233: Implement a good layered log system

valentinbreiz:
Closes #233

## What

Replaces ~700 scattered `Serial.*` calls with a unified `Log.*` API backed by a function pointer sink — init-safe, no interface dispatch, zero cost in release builds for debug messages.

## Architecture

**`LogBackend`** — static `delegate*<string, LogLevel, string, void>` pointing at `DirectWrite` (UART). Swap post-init via `LogBackend.SetBackend(...)` to redirect to framebuffer, ring buffer, network socket, VFS file — same call sites everywhere.

**`[Logger]` + source generator** — annotate any `partial` class with `[Logger]` and the generator emits a typed nested `Log` class with the category baked in:
- `Log.Info/Warning/Error/Critical` — always present, constant strings only in release
- `Log.Debug/Trace` — `[Conditional("COSMOSDEBUG")]`, call sites and string arguments stripped entirely in release builds

**`SerialDebug`** — kept as the allocation-free primitive trace layer for pre-GC-init code (GarbageCollector, PageAllocator, SmallHeap, ManagedModule). These files write raw numbers and hex addresses to UART without constructing heap strings — correct for code that runs before the GC is ready.

**`COSMOSDEBUG`** flag — `Directory.Build.props` + `Sdk.props` wired; `make run DEBUG=1` enables verbose output.

## Migrated in this PR

Post-GC-init (full `Log.*` migration):
- `Idt`, `X64InterruptController`, `ARM64InterruptController`
- `InterruptManager`
- `PciDevice`, `PciManager`
- `LibraryInitializer` (HAL), `Kernel`

Pre-GC-init (`Serial` → `SerialDebug`, stripped in release):
- `GarbageCollector`, `PageAllocator`, `SmallHeap`, `ManagedModule`, `ModuleHelpers`

Remaining ~600 `Serial.*` calls (HAL devices, Scheduler, System, Plugs) are follow-up PRs.

## Tested

Booted clean on x64 and ARM64.


src/Cosmos.Kernel.Core.ARM64/Cpu/ARM64InterruptController.cs
```diff
@@ -36,7 +38,7 @@ public void Initialize()
         GIC.SetPriority(GIC.TIMER_NONSEC_PHYS, 0x80);  // Medium priority
         GIC.EnableInterrupt(GIC.TIMER_NONSEC_PHYS);
 
-        Serial.Write("[ARM64InterruptController] ARM64 interrupt system ready\n");
+        Log.Info("ARM64 interrupt system ready");
```
MrBisquit, line 41: @valentinbreiz Would it be useful to keep things like the `ARM64InterruptController`, but as an optional parameter to `Log.Info` (and the others)?

So like `Log.Info(message)` and then an overload `Log.Info(area, message)`?

PR #341, opened 2026-05-05
VFS interface and manager

zarlo:


valentinbreiz:
Hello @zarlo, really cool API for different FS! Just a question, why splitting VFS between System and HAL? Should not everything stay in System? I don't see why HAL layer is needed here, if we are aligned HAL is for hardware specific components no?

zarlo:
every thing in HAL will be used by drivers and every thing in system it how the user interfaces with the vfs

valentinbreiz:
> every thing in HAL will be used by drivers and every thing in system it how the user interfaces with the vfs

So maybe only the namespace is ambiguous? I see VFS in HAL maybe it should be something like VDriver or something else?

valentinbreiz (approved):

PR #356, opened 2026-05-14
Test runner: per-suite QEMU profile matrix

valentinbreiz:
## Summary
- Per-suite QEMU profile matrix: a suite declares `<CosmosTestProfile/>` + `<CosmosTestModifier/>` items in its csproj and the engine runs the kernel once per matrix cell, prefixing test names with `[profile]` / `[profile+modifier]`. Global catalog in `tests/profiles.json`.
- Two modifier flavors:
  - `deviceOptions` — per-disk-kind QEMU `-device` overrides (e.g. `nvme: msix_qsize=1`).
  - `machineOptions` — `-M` property overrides (e.g. `gic-version=2`).
  - plus optional `architectures` filter so arm64-only knobs vanish on x64.
- New `TR.RunIf(condition, name, action, skipReason)` so tests declare feature requirements; missing-feature cells report as **Skipped** instead of Failed.
- CI workflow renders a per-matrix-cell breakdown in the PR comment alongside per-arch totals — failures attribute to the exact cell.
- Storage suite opts into `{ahci, nvme} × {msix-min, gicv2, gicv3, acpi-off}` for a 5-cell x64 matrix / 9-cell arm64 matrix. New `Cosmos.Kernel.Tests.Pci` suite covers the PCI subsystem (no profiles).
- Defensive engine hardening: `UartMessageParser` drops `TestStart` frames whose decoded name contains control bytes (false-positive magic matches when a kernel IRQ-handler `Serial.WriteString` interleaves with the binary protocol stream), and clamps any `durationMs` > 5 min — kernel-side too.

## Validation
- Green path: 95/95 on x64, 79 pass + 90 skip + 0 fail on arm64 (Storage matrix).
- Failure-attribution flow validated end-to-end via the deliberate `CIValidation_DeviceMustBeSATA` smoke test on `f410552c` (reverted in `edf2322b` once confirmed).

## Out of scope
- Real fix for the `[SCHED] Tick` serial interleave (kernel-side serial lock around `SendMessage` or silencing scheduler tick prints during a test). Engine-side guards above keep the test report clean meanwhile.

## Test plan
- [ ] CI runs the matrix on both arches and the Storage PR comment renders per-profile rows.
- [ ] Failing cells attribute to the exact `[profile]` bucket, not the whole suite.
- [ ] Other suites (HelloWorld, Memory, Pci, etc.) keep single-row-per-arch behaviour.

### DIFF
--- a/tests/Cosmos.TestRunner.Framework/TestRunner.cs
+++ b/tests/Cosmos.TestRunner.Framework/TestRunner.cs
@@ -127,6 +127,21 @@ namespace Cosmos.TestRunner.Framework
         }
 
         /// <summary>
+        /// Run a test that adapts to a capability instead of skipping. The
+        /// <paramref name="condition"/> is passed into <paramref name="test"/>
+        /// so the body can assert the capable path when true and the fallback
+        /// path when false — both branches stay in the report as a real run.
+        /// Use this when the cell always has something to assert but the
+        /// expected outcome differs by profile (e.g. NVMe MSI-X vs polled),
+        /// as opposed to <see cref="RunIf(bool, string, Action, string)"/>
+        /// which reports a Skip when the feature is simply absent.
+        /// </summary>
+        public static void RunIf(bool condition, string testName, Action<bool> test)
+        {
+            Run(testName, () => test(condition));
+        }
+
+        /// <summary>
         /// Run a destructive test whose action is expected to never return
         /// (e.g. a successful Power.Reboot / Power.Shutdown). The test is
         /// pre-emptively reported as passed before invoking the action; if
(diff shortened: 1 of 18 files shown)


zarlo:
@valentinbreiz when this is merge in can you give me a ping, i have been thinkinh off addeding device profiles to the Qemu integration so it will make the VM as close as we can get it to a given device

Guillermo-Santos:
> @valentinbreiz when this is merge in can you give me a ping, i have been thinkinh off addeding device profiles to the Qemu integration so it will make the VM as close as we can get it to a given device

Having device presets would be really cool

valentinbreiz:
> > @valentinbreiz when this is merge in can you give me a ping, i have been thinkinh off addeding device profiles to the Qemu integration so it will make the VM as close as we can get it to a given device
> 
> Having device presets would be really cool

Raspberry Pi preset omg i wait for it (still never tried booting cosmos on rpi)

tests/Cosmos.TestRunner.Framework/TestRunner.cs
```diff
@@ -75,6 +105,27 @@ public static void Run(string testName, Action testAction)
...
+        {
+            if (condition)
+            {
+                Run(testName, testAction);
+            }
+            else
+            {
+                Skip(testName, skipReason);
+            }
+        }
```
Guillermo-Santos, line 127: What do you think about a variant that takes an `Action<bool>`?

valentinbreiz: added it 👍 there is a RunIf(condition, name, Action<bool>) overload that always runs and passes the bool in, so the test asserts the capable path vs the fallback instead of skipping. using it for the nvme msi-x vs polled assertion (nvme+gicv3 must come up msi-x, nvme+gicv2 must fall back to polled)

valentinbreiz:
> @valentinbreiz when this is merge in can you give me a ping, i have been thinkinh off addeding device profiles to the Qemu integration so it will make the VM as close as we can get it to a given device

Will be done in https://github.com/valentinbreiz/nativeaot-patcher/issues/359

PR #358, opened 2026-05-26
Test Runner: Integrate with Microsoft.Testing.Platform

Guillermo-Santos:
This PR tries to integrate Cosmos Test Runner with [Microsoft Testing Platform (MTP)](https://learn.microsoft.com/en-us/dotnet/core/testing/microsoft-testing-platform-intro) for native testing with the `dotnet test` command and VS and VSCode debuggers.

Current Workflow Steps:

1. Build the Test Project with `COSUnit_BuildKernel` empty or set to false.
2. A kernel is Source Generated for each `TestClass` with at least one valid `TestMethod` (Withour parameters and the `Skip` Attribute).
3. Pass the project file as a parameter with --projectfile "$(MSBuildProjectFullPath).
4. Once the Test App Starts it takes the project file and starts the Test Runner Engine, it sets `COSUnit_BuildKernel` to true and `CosmosKernelClass` to the Generated Kernel class, any method marked as `Skip` is reported as `Skipped` as the source generator doesn't see them. 
5. It Collects the results from the Test Runner Engine and translate them to MTP.

# TODO:
- [x] Generate a Kernel from the attributed methods. 
- [x] Build and Execute a Kernel Succesfully.
- [x] Translate results from the Test Runner Engine.
- [x] Migrate the complete Test Runner to Cosmos.TestingFramework, this would allow to do some enhancements to the engine, like live reporting of a test state and Custom Properties for Test Nodes.
- [x] Use testconfig.json for Test Runner Configuration.
- [x] Decide, should one kernel be made for each `TestClass` or only one for all `TestMethods`.
- [ ] Enhance reporting of test duration from the Test Runner Engine, we should Collect Start and End TimeStamps (Both may be collected from the Test Start and Test Success Events respectively if live reporting of Test State is implemented).
- [ ] Refactor current Test Kernels.
- [x] Integrate with VS and VSCode Test Explorers.

# Known Issues: 

1. ~Cosmos Test Runner depends on the project file for building the Kernel, this bring problems when integrating to Visual Studio and Visual Studio Code Test Explorers as they execute the output exe resulted from build, so it doesn't see the project file we append with `TestingPlatformCommandLineArguments`. Consider building a workflow that starts from the Patcher Step downward from the dlls present in the output folder (Current Working Directory for this instances).~ Solved by saving the project full file path into the Assembly Metadata.

# References: 
- [Create custom extensions for Microsoft Testing Platform](https://learn.microsoft.com/en-us/dotnet/core/testing/microsoft-testing-platform-architecture).


Directory.Build.props
```diff
@@ -16,7 +16,9 @@
     <!--pack as nuget-->
     <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
     <!--pack as nuget-->
-
+    
+    <!-- Test Projects should now use MTP -->
+    <UseMicrosoftTestingPlatformRunner>true</UseMicrosoftTestingPlatformRunner>
```
Guillermo-Santos, line 21: This worked locally 😭

Guillermo-Santos:
All all tests, including kernel tests, moved to MTP successfully 🥳

Next thing to do is to add support to the coverage options.

PR #345, opened 2026-05-07
Add FAT12/16/32 driver, EBR + PartitionManager, Vfs format API

valentinbreiz:
## Summary

- Ports a full FAT12/FAT16/FAT32 driver onto zarlo's new Vfs (superblock / inode / file_operations); the driver lives under `Filesystems/Fat/` and registers with `VfsManager`. Future filesystems drop in next to it with no core changes.
- Extends the storage stack so partitions can be created, deleted, resized, and moved across MBR / GPT / EBR through one `PartitionManager` facade. `EBR.cs` walks the extended-partition chain and exposes Add / Remove / Resize / Move; first-deletion promotes the successor into the fixed first slot. `MoveWithData` does direction-aware sector copy before rewriting the table.
- Adds `IVfsFilesystemType.TryFormat / TryDestroy` (with a per-driver `IVfsFormatOptions` marker) so any registered filesystem can be mkfs'd through `VfsManager`. The FAT driver implements both via `FatFormatter` (BPB + all FAT copies + FSInfo / backup boot for FAT32 + zeroed root area; SPC / reserved / FAT-size auto-pick when not specified).
- `CosmosEnableFat` MSBuild flag (default `true`, cascades from `CosmosEnableStorage`); `CosmosFeatures.FatEnabled` exposed via `FeatureSwitchDefinition` for ILC trimming.

## Tests

- **FAT suite (new): 26/26 passing on x64 in 96 s.** Setup × 6, FAT16 inode-level × 6, FAT32 + Vfs-handle × 11 (incl. LFN, 256 KiB across 512 clusters, root growing across clusters, persistence across unmount), plus `Test_Vfs_Format_Fat32_RoundTrip`, `Test_Vfs_Destroy_PreventsRemount`, and `Test_Vfs_Format_Fat12_Mountable` covering the new format/destroy API end-to-end.
- **Storage suite: 78 tests** covering full create / delete / move / resize lifecycle for MBR, GPT, and EBR via `PartitionManager`, including overlapping-forward `MoveWithData` and logical-partition moves.

## Test plan

- [x] `make test KERNEL=Fat ARCH=x64` (26/26 passing)
- [ ] `make test KERNEL=Fat ARCH=arm64`
- [ ] `make test KERNEL=Storage ARCH=x64` to confirm no regression from EBR additions
- [ ] `make test KERNEL=Storage ARCH=arm64`
- [ ] DevKernel boots with FAT registered (`TryRegisterFat`) and attempts mount on partition `0`

PR #366, opened 2026-07-10
Feature/fat vfs

valentinbreiz:
See https://github.com/valentinbreiz/nativeaot-patcher/pull/345


Demiomad:
does this mean what i think it means...?

PR #381, opened 2026-07-17
DNS: resolve CNAME chains and support multiple A records

Cristiandis:
- Add RFC 1035 compression-pointer decoding (ParseNameAt) so resource record names and CNAME RDATA can be decompressed instead of only raw pointer bytes
- Add DNSAnswer.ResolvedName and DNSAnswer.CanonicalName, populated during DNSPacketAnswer parsing
- Add DNSRecordType (A, CNAME) constants
- DnsClient: follow CNAME chains by name instead of matching raw bytes, with loop detection against malformed/looping responses
- Add DnsClient.ReceiveAll() returning every A record for the resolved name; Receive() now returns the first of these, preserving its existing signature/behavior
- Fix message-base offset used when resolving compression pointers: the DNS message (and pointer offsets) start at DataOffset + 8, not DataOffset, since the 12-byte header begins at the Transaction ID field written at that offset


valentinbreiz:
Thanks for the PR!

valentinbreiz:
Format CI is failing, a `dotnet format ./nativeaot-patcher.slnx` is needed

Issue #357, opened 2026-05-14
Threading: latent heisenbug on arm64 — thread-exit resume from `_monitor.Wait` (masked by diagnostic traces)


valentinbreiz:
## Background

During the work on #355 (live debug snapshot buffers), the `threading-tests / Threading Tests - arm64` job started failing intermittently with a hang at test 16 (`SpinLock_ProtectsSharedData_AcrossThreads`). Test 16 was the first failing test; the remaining 33 tests then timed out behind it.

Initial failing run UART: `gh run download 25864902710 --repo valentinbreiz/nativeaot-patcher --name uart-log-threading-arm64`.

## What we found

In the failing UART the sequence after thread 6 (`SpinLockThread1Worker`) finishes is:

```
[SCHED] Thread 6 exiting with code 0
[ThreadPlug] Invoking managed thread exit callback for thread 6
[SCHED] Thread 6 is now ready, RSP=0x59B6BAC0       ← from ConditionVariable.Signal
[ThreadPlug] CreateThread(GCHandle<Thread>)         ← idle reached thread2.Start()
...
[ThreadPlug] Invoking managed thread exit callback for thread 7
[SCHED] Tick 950 enabled=1                          ← no "OnThreadExit done" ever
```

`is now ready` is only emitted by `SchedulerManager.ReadyThread`, which on this path can only be reached via `ConditionVariable.Signal/SignalAll`. So thread 6's exit callback entered the wait subsystem:

`OnThreadExit` (Thread.NativeAot.Unix.cs) → `_stopped.Set()` → `WaitSubsystem.SetEvent` → `s_lock` (`LowLevelLock`) contended → `_monitor.Wait()` → `InteropSysPlug.LowLevelMonitor_Wait` → `Monitor.Wait` → `ConditionVariable.Wait` → `SchedulerManager.BlockThread`.

`s_lock` is contended because the main idle thread is in `Thread.StartCore`'s spin loop (`while (Unstarted && !JoinInternal(0)) Yield();`), where `JoinInternal(0)` repeatedly acquires/releases `s_lock`.

[...]

PR #400, opened 2026-07-21
Add tick-driven software timers and a thread-context AlarmSystem

valentinbreiz:
Adds two ways to run code after a delay: `SoftwareTimer`s driven by the hardware tick of any timer device (interrupt context), and `AlarmSystem`, an alarm service running callbacks on a dedicated kernel thread (thread context).

---

## Why

The only deferred-callback mechanism was `PIT.PITTimer`, locked inside the x64 PIT driver — ARM64 had no equivalent, and callbacks could only ever run in interrupt context, where they must not block. There was no way to say "run this code in X ms" portably, and no way at all to do it from code that needs to sleep, allocate freely, or take locks.

## What changed

| | `TimerManager.Schedule` / `ScheduleRecurring` | `AlarmSystem.Add` / `AddRecurring` |
|---|---|---|
| Callback context | tick interrupt handler | dedicated kernel thread |
| May block / take locks | no | yes |
| Resolution | hardware tick (~55 ms PIT, 10 ms Generic Timer) | scheduler tick (~10 ms) |
| Needs scheduler | no | yes |
| Cancel | `Cancel(timer)` | `Remove(id)` |
| Lives in | `Cosmos.Kernel.System.Timer` | `Cosmos.Kernel.Core.Scheduler` |

### SoftwareTimer (interrupt context)

`SoftwareTimer` (one-shot or recurring, nanosecond timeout) moves the old `PITTimer` concept into `Cosmos.Kernel.HAL.Interfaces`, with the registry in the shared `TimerDevice` base: `RegisterTimer`/`UnregisterTimer`/`HandleTick(elapsedNs)`. The PIT (IRQ 0) and the ARM64 Generic Timer (PPI 30) both feed their tick into `HandleTick`, so the same timer works on both architectures. `TimerDevice.Wait` is now a shared implementation on top of it (one-shot timer + halt loop); `GenericTimer` keeps its counter-poll override. Registration is IRQ-atomic; unregistration scans by reference to keep `EqualityComparer<T>.Default` out of kernel paths.

`PIT.PITTimer` and `PIT.WaitNS` are deleted — nothing in-tree used them.

### AlarmSystem (thread context)

Alarms are absolute Stopwatch-tick deadlines in a list sorted by due time. A lazily started kernel thread pops due alarms and parks in `ConditionVariable.WaitTimeout` until the next deadline. `Add` inserts under the same mutex the wait atomically releases, so a new earlier alarm either lands before the deadline computation or finds the thread parked and signals it — the CV's atomic insert+release+park (#357 fix) closes the lost-wakeup window.

Callbacks fire outside the lock, so they may block and call `Add`/`Remove` themselves. Recurring alarms re-arm from now rather than from the missed deadline (no catch-up bursts), and callback exceptions are caught and logged instead of killing the service thread. Without a running scheduler, `Add` returns 0 and logs.

## Tests

The Timer suite grows ten tests, all running on both architectures (`expectedTests` 14→24 on x64, 8→18 on ARM64). Per layer (SoftwareTimer, AlarmSystem): a one-shot fires exactly once and deactivates, a 50 ms recurring timer fires ≥3 times in 500 ms, and a cancelled/removed timer never fires. A last group covers the BCL `System.Threading.Timer` (one-shot, periodic, dispose-cancels) and `Task.Delay` — the first TimerQueue coverage in a kernel, viable now that `MarkSleeping` computes accurate timed waits.

Validated by the CI kernel suites on both architectures; the first CI round surfaced three pre-existing scheduler bugs, fixed here (see the investigation comment below).

## Notes for review

- The alarm thread's waits go through `ConditionVariable`, which logs `[CV]` debug lines on every wait/signal — chatty with recurring alarms. Left untouched here; stripping those logs is a separate cleanup.
- On x64 the software-timer tick is the PIT at its ~55 ms default period (the LAPIC drives scheduling), so `TimerManager.Schedule` granularity is coarser on x64 than ARM64 until the PIT period is raised.


valentinbreiz:
The first CI round failed the Timer suite on both architectures (x64 hang at `AlarmSystem_Add_Fires`, ARM64 12/14 with both alarm fire counts at 0). All three root causes are pre-existing scheduler bugs that these are the first tests to exercise — fixed in 5fb4e938 and 84c9252b:

| Bug | Evidence in the UART logs | Effect |
|---|---|---|
| `LocalApic.Wait` masks the LVT for its one-shot polling wait and never restores periodic mode | LAPIC ticks 1–5 log at boot; the unconditional tick-100 log never appears | Scheduler tick permanently dead on x64 after the `LAPIC_Wait` tests: no preemption, no quantum accounting, no `CheckSleepingThreads` |
| `BlockThread` never sets `NeedReschedule` (unlike `ReadyThread`) | 186 `BlockThread id=1` lines, one per PIT IRQ, main thread silent | With the tick dead, the alarm thread re-entered its block-halt loop forever while the mutex owner (the idle thread, preempted mid-`Add`) never ran again |
| `MarkSleeping` adds `timeoutMs * 1_000_000` to `GetTimestamp()` — nanoseconds onto Stopwatch ticks | Alarm thread parks with `timeoutMs=40`, never times out, only a later `Signal` wakes it | Timed waits stretched 16× on ARM64 (62.5 MHz generic timer): a 40 ms park slept 640 ms, so no alarm fired inside the 500 ms test window; on x64 (multi-GHz TSC) timeouts shrank ~2.6× instead |

The idle thread masked the `MarkSleeping` bug for `Thread.Sleep`: it gets resurrected every tick by the `PickNext ?? IdleThread` fallback regardless of its deadline, so its sleep loops re-check wall-clock time and still add up correctly. A parked non-idle thread has no such safety net.

Related observation, not addressed here: `StrideScheduler` mixes the same units (`LastWakeup`/`GetTimestamp` ticks vs `WakeupBoostDecayNs`, pass increments against `DefaultQuantumNs`), which skews the interactive-boost heuristics but doesn't break correctness.

PR #404, opened 2026-07-22
refactor: makes immutable and optimizes Address class

MihaMarkic:


MihaMarkic:
Shall I also go ahead and remove IComparable as Equals is more than enough?

zarlo:
IComparable can be useful

PR #406, opened 2026-07-23
refactor: optimizes Tcp and byte buffers handling by using shared pools

MihaMarkic:
Optimizes Tcp:
- uses ArrayPool<byte> for buffers
- makes Data access private
- Tcp instances are created only through static Tcp.CreateConnection method, so they can be tracked and disposed eventually to release rented buffers
- introduces unit tests


MrBisquit (changes requested):
Need to run `dotnet format`

PR #408, opened 2026-07-24
feat: GC Interior Pointers and Other Enhancements

### LINKED ISSUE #376: GC: Add support for interior pointers
When an object is identified as an interior pointer (GcCallFlags.GC_CALL_INTERIOR is present) then we should try and find the parent object and mark it instead of only the child.

We already have the basis for implementing this in [GarbageCollector.PreciseStack.cs](https://github.com/valentinbreiz/nativeaot-patcher/blob/main/src/Cosmos.Kernel.Core/Memory/GarbageCollector/GarbageCollector.PreciseStack.cs) it even is a parameter of the `PreciseRootTrampoline` method that is the one that calls `TryMarkRoot`.

Either add the gcCallFlags as a parameter of `TryMarkRoot` or handle it directly on `PreciseRootTrampoline`, if the latter is selected then it should search for the parent object and send it to `TryMarkRoot` instead of the child object.

This could start as a simple slow search where we compare our object pointer first against all segments, to see in what segment it is contained, and then to each object on the segment till we find the one where our pointer is inside, but we may need a way to improve this speed (see reference for details).

# References

- [Writing a .NET Garbage Collector in C#  - Part 8: Interior pointers](https://minidump.net/writing-a-net-gc-in-c-part-8/)

### LINKED ISSUE #384: GC: interior pointers are dropped as roots — objects referenced only through a byref/Span are collected while in use
## Symptom

An object whose only live reference is an interior pointer (a `ref` into its fields/elements, a `Span<T>`'s `_reference`) is collected by the GC while still in active use. Once the freed block is recycled, `AllocFromFreeList` zeroes it — the still-held byref then reads zeros or another object's data.

`StringBuilder.Append` works through a `Span<char>` over the current chunk; when ILC dead-codes the chunk/builder reference after its last use, the span's interior pointer is the only root left.

## Proof

DevKernel repro (framework at 5dff4549):

```csharp
[MethodImpl(MethodImplOptions.NoInlining)]
private static ref int LeakInteriorRef()
{
    // TLAB-sized (8416 bytes) so that, if wrongly freed, the churn's TLAB refill
    // takes the free block via AllocFromFreeListRaw(8192) and MemSets it to zero.
    int[] arr = new int[2100];
    for (int i = 0; i < arr.Length; i++) { arr[i] = 0x5A5A0000 + i; }
    return ref arr[8];
}

int[] control = new int[8];      // normal reference held across the collection
control[0] = 0x0C0C0C0C;

ref int r = ref LeakInteriorRef();   // byref is now the ONLY root of the array
Heap.Collect();
for (int i = 0; i < 512; i++) { int[] junk = new int[32]; junk[0] = i; }  // churn
```

Serial output:

```
[InteriorProof] before collect: 5A5A0008
[InteriorProof] after collect:  5A5A0008   (freed, not yet recycled)
[InteriorProof] after churn:    0          <-- array was collected and zero-recycled
[InteriorProof] control:        C0C0C0C    <-- normal reference survives
```

## Root cause

`TryMarkRoot` (`GarbageCollector.Mark.cs`) only roots a value if it lands exactly on an object header — it reads the MethodTable at the pointed-to address and drops anything that fails validation. A pointer into the middle of an object always fails. Both scanners funnel through it:

- **Precise scan**: `PreciseRootTrampoline` (`GarbageCollector.PreciseStack.cs`) receives slots the GcInfo explicitly flags `GC_CALL_INTERIOR` and passes them to `TryMarkRoot` as-is. The comment already documents the hole: *"An interior pointer (GC_CALL_INTERIOR) is passed through as-is — TryMarkRoot's MethodTable check then drops a non-header byref, the same hole the conservative scanner has."*
- **Conservative scan** (parked threads): any stack word pointing into an object's interior is dropped the same way.

It usually goes unnoticed because the object-start reference is typically also still live somewhere; it bites exactly when codegen kills the object reference after its last use while a byref/span remains — the `StringBuilder.Append` pattern.

## Proposed fix

Add `FindContainingObject(nint ptr)` that resolves an interior pointer to the containing object's start — locate the segment, walk it with the sweep's logic (free block → skip `Size`, filler → skip 8, object → skip `ComputeSize`) until the range containing `ptr` is found — then mark that object:

- Precise scan: resolve only slots flagged `GC_CALL_INTERIOR` (exact, cheap).
- Conservative scan: when the exact-header check fails but the value points into the GC heap, resolve containment (CoreCLR-conservative semantics; trades some floating garbage for soundness).
- Same resolution for pinned segments; medium/large heaps resolve via their page headers.

Cost is only paid during GC scans. The repro above is a ready-made kernel test: hold only a byref across a forced `Collect()` + churn and assert the contents survive.

Guillermo-Santos:


src/Cosmos.Kernel.Core/Memory/GarbageCollector/GarbageCollector.PreciseStack.cs
```diff
@@ -162,6 +162,51 @@ private static void PreciseRootTrampoline(void* ctx, nuint* pObjRef, uint gcRefF
...
         TryMarkRoot((nint)(*pObjRef));
     }
+
+    private static void* GetParentObject(void* obj)
+    {
+        var segment = s_segments;
+
+        while (segment != null)
+        {
+            if (segment->Start >= obj && segment->End < obj)
```
Guillermo-Santos: 3AM Code be like (the operands are inverted lol).

valentinbreiz (approved):

PR #436, opened 2026-08-01
Honor Thread constructor maxStackSize via SystemNative_CreateThread

### LINKED ISSUE #435: Thread constructor maxStackSize is ignored
`ThreadPlug.CreateThread` always allocates `DefaultStackSize` (256KB since #434); the `maxStackSize` passed to `new Thread(start, maxStackSize)` lands in `_startHelper._maxStackSize` and upstream's `CreateThread` honors it, but our plug never reads it.

With stacks committed upfront (no demand paging), kernels that spawn many small workers have no way back to a leaner footprint. Honoring the constructor gives them one without touching the default: `new Thread(start, 64 * 1024)`.

Proposed semantics, mirroring the dotnet sources:

- `<= 0` → `DefaultStackSize` (matches `Thread.NativeAot.Unix.cs`)
- otherwise page-align up and floor at 64KB — CoreCLR's "arbitrary minimum" for stack-size settings (`RhConfig.cpp`)
- no 128KB clamp: a sub-128KB request means `EnsureSufficientExecutionStack` throws on that thread, same as upstream Linux honoring a small pthread stack

Suggested by @Guillermo-Santos after #434 raised the default.

valentinbreiz:
Fixes #435

## Problem

`new Thread(start, maxStackSize)` was silently ignored: `ThreadPlug.CreateThread` always allocated `DefaultStackSize`. With stacks committed upfront (no demand paging), kernels spawning many small workers had no way back to a leaner footprint after #434 raised the default to 256KB.

The fix — reading `_startHelper._maxStackSize` from the plug — is impossible: `StartHelper` is a private nested type, and `UnsafeAccessor` field accessors with an `[UnsafeAccessorType]` byref return are specified to throw `NotSupportedException` (`Verify_Type_GetInstanceFields_NotSupported` in the runtime's own tests; confirmed live as a kernel crash).

## Fix

Let upstream do the reading and plug the seam below it instead:

- `Thread.CreateThread` is no longer plugged. The upstream body legally reads the private `StartHelper`, resolves `<= 0` to `RhGetDefaultStackSize`, and passes the result to `Interop.Sys.CreateThread`.
- Core exports the two FCalls that body needs: `RhGetDefaultStackSize` (returns `DefaultStackSize`, 256KB) and `RhGetThreadEntryPointAddress` (unused address — Cosmos threads start at `ThreadNative.EntryPointStub` and the scheduler runs CoreLib's `StartThread` with the GCHandle parameter itself).
- Core's `libSystemNative` exports `SystemNative_CreateThread` — the old plug body, now receiving the resolved stack size as a plain parameter. It page-aligns the request and floors it at 64KB, CoreCLR's arbitrary minimum for stack-size settings (`RhConfig.cpp`), standing in for the `PTHREAD_STACK_MIN` clamp in upstream's `pal_threading.c`. The export lives in Core because only Core is in `UnmanagedEntryPointsAssembly` (`--generateunmanagedentrypoints`).
- `InteropAPI.txt` gains `libSystem.Native!SystemNative_CreateThread`. Without the direct-link entry the P/Invoke compiled as a lazy import whose stub returned garbage "success", so `StartCore` spun forever waiting for a thread that was never created.

Requests below CoreLib's 128KB `MinExecutionStackSize` are honored, matching upstream on Linux: `EnsureSufficientExecutionStack` throws on such a thread by contract. Sub-128KB workers are therefore for code that stays off record `ToString` and friends.

## Reproduction and verification

Two new Threading cells: the worker reads its own scheduler thread's `StackSize`. `Thread_MaxStackSize_IsHonored` requests 512KB and expects exactly that plus a passing `TryEnsureSufficientExecutionStack`; `Thread_MaxStackSize_TinyRequestIsFloored` requests 16KB and expects the 64KB floor plus `TryEnsureSufficientExecutionStack` reporting false.

Serial log confirms the sizes end-to-end: default threads log `stack 262144 bytes`, the probes log `524288` and `65536`.

| Run | Result |
| --- | --- |
| Threading x64 | 58/58 |
| Threading arm64 | 58/58 |
| GarbageCollector x64 | 44/44 |

### DIFF
--- a/src/Cosmos.Kernel.Core/Bridge/Interop/libSystemNative.cs
+++ b/src/Cosmos.Kernel.Core/Bridge/Interop/libSystemNative.cs
@@ -1,10 +1,12 @@
 using System.Diagnostics;
 using System.Runtime.InteropServices;
+using Cosmos.Kernel.Core.CPU;
+using Cosmos.Kernel.Core.IO;
 using Cosmos.Kernel.Core.Scheduler;
 
 namespace Cosmos.Kernel.Core.Bridge.Interop;
 
-internal static class libSystemNative
+internal static unsafe partial class libSystemNative
 {
     [StructLayout(LayoutKind.Sequential)]
     internal struct ProcessCpuInformation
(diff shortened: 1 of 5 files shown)


Guillermo-Santos (approved):
👌

PR #451, opened 2026-08-23
Decide the public surface before the first release freeze

valentinbreiz:
Last work stream of the pre-release API audit. #447 built the tracking tooling and #452 documented the surface it declared, but neither decided what belongs on that surface in the first place. This branch states the policy, applies it across all four tracked assemblies, and proves the extension points it opens by writing a second consumer for each.

It absorbed the rest of the audit as it went: #452 (the XML doc pass), #454 (`Canvas3D`) and #460 (the ring policy and the scheduler seam) are merged in, so this is 74 commits rather than the "part 1" the title promised. The streams are called out separately below.

## Problem

Visibility was the extension mechanism. `public` meant "another assembly in this repo needed it", so the packages shipped their whole implementation as API:

| Symbol | Issue |
|---|---|
| `ICSharpCode.SharpZipLib.*` (12 types) | Vendored third-party code fully public, inconsistent with BigGustave and LunarLabs.Fonts (internal) |
| `Conversion.Base64` | Single undocumented unsafe API `byte* Decode(char*, uint)`, zero callers in the repo |
| `LibraryInitializer` (four copies) and `EagerStaticClassConstructionAttribute` (two of its three copies) | Public only because NativeAOT startup calls them by convention; tracked as if they were offered API |
| `OutgoingBuffer.BufferEntry.EntryStatus` | Public enum nested in a private class, unreachable |
| `Math.cs`, `Graphics/Color.cs` | Empty files |
| `IVfsDirectoryHandle` | Not `IDisposable` while `IVfsFileHandle` is |
| The wire-format packet classes, the TCP machinery (`Tcp`, `TransmissionControlBlock`, `Flags`, `Status`, `OutgoingBuffer`) and `UdpClient.rxBuffer` | All equally public, with nothing marking which were meant as a contract and which were reachable by accident |
| `ConsoleStream`, `ConsoleTextWriter`, `KeyboardTextReader` | Plumbing behind `System.Console`, only consumed by the console plugs |
| `Cell` + `KernelConsole.GetCellAt`/`SetCellAt`, `UnicodeMapping` + the mapping-list `PCScreenFont` ctor | Renderer internals, no consumer outside the assembly |
| FAT on-disk layer (`FatTable`, `FatDirectory` + 13 layout consts, `FatAttr`, `FatDirEntry`) | Directory-encoding engine used by the inode operations; only external consumer is the white-box FAT test suite |
| `Kernel.mStarted`/`mStopped`, `KeyMapping` (13 fields), `NetworkConfigEntry`, `EndPoint`, `Image.RawData`, `DNSConfig.DNSNameservers`, `NetworkConfigManager.NetworkConfigs`, `UdpClient.DynamicPortStart` | Public mutable fields (gen2 ports), exposed mutable static lists, gen2 Hungarian naming on the kernel base class |
| Setters on `KeyboardManager` modifier/lock state, `MouseManager.ScreenWidth/Height`, `NetworkConfigManager.CurrentNetworkConfig`, `Global.CurrentKernel` | Driver- or framework-owned state writable by any kernel |
| `DNSConfig`, `DHCPClient`, `ICMPClient` next to `DnsClient` | Mixed acronym casing; .NET style Pascal-cases acronyms of 3+ letters |
| `Mbr.PartitionEntry` and `Gpt.PartitionEntry` | Public nested types, discouraged by the framework design guidelines |
| `Cosmos.Kernel.Core` and `Cosmos.Kernel.HAL` in their entirety | Drivers, PCI, ports, the native bridges, the metadata readers and the NativeAOT convention hooks, all reachable, none of it intended as a contract, and none of it tracked |

Tracking a surface that nobody had decided on would only have frozen the accident.

## The rule

Three rules, written into [Public API Tracking](docs/articles/dev/public-api.md) and applied across the tree:

1. **One supported ring.** `Cosmos.Kernel.System` is what kernels program against: tracked, documented, frozen at release, with deprecation cycles applying there and only there. A contract type from another assembly rides along only where a kernel can obtain one or meaningfully supply one through the ring. `IBlockDevice` in `HAL.Interfaces` is the one device contract that passes, and it passes on both halves: `StorageManager.PrimaryDevice` hands a kernel a device, 34 public members take one, DevKernel reads and writes blocks straight through it, and the Fat and File suites each implement a RAM disk of their own and drive it holding no grant. The VFS contracts in `HAL` pass through `VfsManager.RegisterFilesystem` and the handle API. Nothing in `Core` passes, so Core's whole public surface is the scheduler seam: all 81 of its tracked entries carry the `COSMOS0001` prefix.

   The weaker version of this test, "a ring signature names it", is what the first pass used, and it is wrong in both directions: it passes anything somebody once wrote a `Register` overload for and fails an identical contract nobody did. Two sections below trace where it put `IKeyboardDevice` and `INetworkDevice`.
2. **Chosen experimental seams.** Extension points are opened deliberately under `[Experimental]`: usable now, no compatibility promise, promoted by dropping the attribute. Referencing one is a build error until the consumer suppresses its diagnostic ID, which is the acknowledgement. `COSMOS0001` is the scheduler policy seam, `COSMOS0002` the packet seam.
3. **Everything else is internal.** First-party assemblies and the white-box test kernels reach internals through `InternalsVisibleTo`; anyone else goes through `[UnsafeAccessor]` and accepts that internals move.

The same question applied to the first-party consumers: a facade is only worth having if the code that could use it does. `Cosmos.Kernel.Plugs` holds a Core grant it genuinely needs, for the memory and heap internals behind `GcPlug` and `NativeMemoryPlug`, the runtime internals, the native bridges and `Core.Scheduler.Monitor`, but two of its Core dependencies had a ring route and kept using the internal: `Serial` at 113 call sites, where `Log` is a stateless 1:1 forwarder and the test framework had already moved, and `CosmosFeatures` at four, where `KernelFeatures` forwards to the same flags. Both moved. No plug used `AlarmSystem`.

The enforcement test is mechanical rather than a matter of taste: `examples/DevKernel` must compile with no `InternalsVisibleTo` grant. If DevKernel needs a symbol, the symbol becomes public or earns a `Cosmos.Kernel.System` facade; if it does not, it stays internal. That forced the facades this branch adds — logging, memory, scheduler stats and feature switches — and the ports of DevKernel, the black-box test kernels and the test framework onto them.

## Fix

**The System sweep.** SharpZipLib top-level types flipped to `internal` (visibility-only diff, stays diffable with upstream; the only consumer is the Png decoder in the same assembly). `Base64.cs` deleted along with the empty `Conversion/` folder, `Math.cs` and `Color.cs`. `EntryStatus` made internal. `IVfsDirectoryHandle : IDisposable` with a no-op `Dispose` on the handle. All the plumbing in the table above made `internal`, with `Cosmos.Kernel.Plugs`, `Cosmos.Kernel.Tests.Fat`, `Cosmos.Kernel.Tests.Network` and `Cosmos.Kernel.Tests.Graphic` keeping access through `InternalsVisibleTo`, next to the pre-existing `Cosmos.Kernel.Tests.System` grant (patched plug bodies referencing another assembly's internals compile and run fine: ILC does not enforce IL-level accessibility, confirmed by the Network suite below).

Kept public on purpose: `FatBootSector` (partition probing in StorageView/StorageManager); `ColorOrder` (parameter of `Bitmap`'s public constructors); the keyboard layouts and `KeyMapping`/`ScanMapBase` (documented custom-layout API).

Field-to-property pass: `Kernel` lifecycle flags become `Started`/`Stopped` protected properties, `KeyMapping`/`NetworkConfigEntry`/`EndPoint`/`Image.RawData` become properties, `DnsConfig.Nameservers` becomes an `IReadOnlyList` view (Add/Remove stay the mutators), `NetworkConfigs` and `DynamicPortStart` go private, and the state setters above go private (`MouseManager.Sensitivity` stays writable, it is the documented knob). Acronym casing normalized to `DnsConfig`, `DhcpClient`, `IcmpClient` and the Tcp/Udp/Arp/Dhcp/Dns packet types, files renamed to match; `IPConfig`/`IPAddress`/`IPPacket` stay (two-letter acronyms remain uppercase in .NET style), `Flags` becomes `TcpFlags` so the seam's own type does not read as a bare word, and the namespace segments (`TCP`, `UDP`, `DHCP`, `DNS`, `ARP`) are untouched because `IPv4.Tcp` would collide with the `Tcp` class. `Mbr.PartitionEntry`/`Gpt.PartitionEntry` hoisted to top-level `MbrPartitionEntry`/`GptPartitionEntry`.

DevKernel rewritten off the raw-packet anti-pattern: `netsend` used to build a `UDPPacket` with a broadcast MAC and push it straight into the NIC behind a stale "the stack does not resolve ARP yet" comment; it now uses `UdpClient` Connect/Send. `netlisten` used the promiscuous `UDPPacket.OnUDPDataReceived` hook; it now receives on a `UdpClient` bound to the test port and stops on Esc.

**Core and HAL.** This is the source-breaking half, and it was not in the original scope. Both assemblies are internalized down to their contract types — 92 type declarations across 72 files in Core, 142 across 123 files in HAL — and then newly tracked, alongside `HAL.Interfaces`. What stays public in Core is the scheduler seam and nothing else; in HAL it is the VFS contracts, and in `HAL.Interfaces` the device contracts. All `[LibraryImport]` native bridge classes went internal (Core's 8, Core.X64's 5, Core.ARM64's 8, plus `ManagedModule` and `ExceptionHelper`), with cross-assembly consumers keeping access through `InternalsVisibleTo`. Any out-of-repo kernel touching a Core or HAL type will stop compiling.

The NativeAOT convention hooks went internal with everything else rather than being carved out of tracking: ILC's `--initassembly` lookup does not care about visibility, so all four `LibraryInitializer` copies and the three `EagerStaticClassConstructionAttribute` copies are now `internal`. The metadata readers were vendored into `Core/Runtime/Internal/` for the same reason — they are linked from the runtime submodule, where they are public, and a local copy is what makes them internal here.

**The contracts that were public by accident.** Applying rule 1's own test to the result found 46 tracked entries that nothing outside the repo could call. `Cosmos.Kernel.System`'s 1051 entries name `Cosmos.Kernel.Core` zero times, so the platform interfaces never rode along on the ring at all: they were public because `IPlatformInitializer` returns them, and `IPlatformInitializer` is itself unreachable, since `PlatformHAL` is internal and no kernel can register or obtain an initializer. `IGraphicDevice` is the same shape, with no `GraphicsManager.RegisterDevice` to match the other five managers and an already-internal implementer. `IPortIO`, `ICpuOps`, `IPowerOps`, `IInterruptController`, `IRQContext`, `IPlatformInitializer` and `IGraphicDevice` are therefore internal, taking Core from 100 tracked entries to 81.

The input and timer contracts fail the same test one step later. `IKeyboardDevice` and `IMouseDevice` each have exactly one ring anchor, a `Register` method whose only caller in the tree is System's own `LibraryInitializer` wiring HAL enumeration into the managers at boot; `ITimerDevice` adds only a `TimerManager.Timer` getter with nothing public to do with the result. The difference from `IGraphicDevice` was that somebody wrote `RegisterKeyboard` and nobody wrote `RegisterGraphicDevice`. A kernel reads `KeyboardManager.ReadKey` and `MouseManager.X`, never the device, so those three plus `KeyPressedHandler` and `TimerTickHandler` go internal, and `RegisterKeyboard`, `RegisterMouse`, `RegisterTimer` and `TimerManager.Timer` follow them. `TimerManager.Schedule`, `ScheduleRecurring` and `Cancel` are untouched, since `SoftwareTimer` never mentions `ITimerDevice`.

`INetworkDevice` is where the weak test held out longest. Thirteen ring members take one, which reads as decisive until you count what a kernel does with the handle: it reads `Name`, `MacAddress`, `LinkUp` and `Ready`, and null-checks it. Across DevKernel and the network suite nothing ever called `Enable`, `Disable`, `Initialize` or `Send` on it, all 17 acquisitions were `PrimaryDevice` and none was `GetDevice(index)`, and `RegisterDevice` had exactly one caller, System's own `LibraryInitializer`, the same shape that sent `RegisterKeyboard` internal. `NetworkManager` was already fronting `LinkUp` and `Send` off the primary device, so the facade was half built and callers reached past it. It gains `HasDevice`, `Name`, `MacAddress` and `Ready`; `PrimaryDevice`, `GetDevice` and `RegisterDevice` join the contract as internal, and the stack members that took a device follow: both `NetworkStack.ConfigIP` overloads, the five device-keyed `NetworkConfigManager` members, `IPConfig.Enable`, `DhcpClient.DHCPServerAddress` and `NetworkConfigEntry`. The two a kernel called gain device-free forms, `IPConfig.Enable(ip, subnet, gw)` and a device-free way to read the configuration back, which is how both call sites already used them. `NetworkStack.Update` and `NetworkConfigManager.ClearConfigs` go with them: neither had a caller outside System, and `ClearConfigs` dropped the config list while leaving the stack's address and MAC maps populated.

Hiding the contract must not hide a capability with it, and it did: `IPConfig.Enable(device, ...)` was the only way to configure a chosen NIC. Choosing the *primary* was never possible either, on this branch or before it, since `s_primaryDevice` was assigned once in `RegisterDevice` and `PrimaryDevice` was get-only. Everything below the facade is already multi-device: `DhcpClient` loops every registered device, `OutgoingBuffer` routes each packet by matching its source address through `IPConfig.FindInterface`, and `MACMap` and `AddressMap` are keyed per device. So the ring gains `NetworkAdapter`, a `readonly struct` carrying the registration index, with `NetworkManager.GetAdapter`, a settable `NetworkManager.Primary`, an adapter overload of `IPConfig.Enable`, and `NetworkAdapter.IPConfig` for the configuration in force on one. The ring owns the handle, so the policy is met by a type a kernel can really obtain rather than by omission. The index is stored biased by one so a default-constructed handle names nothing instead of device 0, the `InterruptMaskScope` trap this PR already fixed once, and equality is written out rather than inherited, because `ValueType.Equals` and `GetHashCode` go through reflection. DevKernel's `netinfo` lists the adapters and marks the primary, holding no grant.

That raised a fair question in review: why does `IPConfig` get a `NetworkConfigManager` when `DnsConfig` does not. Measuring it, the manager was not one. Configuring a single device wrote the same config to three stores from `NetworkStack.ConfigIP`: `IPConfig`'s own routing list, read by `FindNetwork`, `IsLocalAddress` and `FindRoute`; the manager's device-keyed list, read by `Get`; and the manager's current pointer, read by `Current` and `CurrentAddress`. They were written together and cleared apart, and the two `Remove` methods that would desynchronize them are both dead code, which is the evidence that the split accreted rather than was designed. There is one store now, device-keyed, inside `IPConfig`, and it replaces a device's entry rather than appending, so a reconfigured device leaves no stale route to drop packets on. The per-device lookup moved onto `NetworkAdapter.IPConfig`, where the ring already keeps `Name`, `MacAddress`, `LinkUp` and `Ready`, and `NetworkConfigManager` is deleted. `Current` and `CurrentAddress` become `NetworkManager.Primary.IPConfig`, which differs only with more than one NIC: `Current` was last-writer-wins, `Primary` is the adapter the kernel chose. `DnsConfig` keeps its own list rather than gaining a manager to match, because a nameserver list is global to the resolver while an IPv4 config belongs to one device; it becomes a `static class`, dropping the phantom public constructor it was carrying.

The test then reruns member by member, because a type earning its place does not carry its whole class in. `SoftwareTimer` stays public, since `TimerManager.Schedule` returns one and `Cancel` takes it back, but its constructor, `SetActive`, `Tick` and `Invoke` do not: every call site is the timer device that owns the registry, and from a kernel each one corrupts it. `MACAddress.bytes` was a public mutable array behind a `readonly` field, and `NetworkStack.MACMap` is keyed on a hash derived from it; it is internal, renamed `_bytes` because `EnforceCodeStyleInBuild` makes the `_camelCase` rule for internal fields a build error.

Reviewing whether `SoftwareTimer` should follow `INetworkDevice` behind its manager turned up the opposite problem one step over. `TimerManager.Schedule` and `ScheduleRecurring` both told kernels, in their public XML docs, to use `AlarmSystem` for callbacks that need thread context, and `AlarmSystem` is internal to `Cosmos.Kernel.Core`. The ring was handing out the interrupt-context half, where a callback must not block, allocate or take a lock, and pointing at the safe half through a wall; the only consumer was the Timer suite, through a Core grant. `AlarmManager` is the ring route, forwarding to the internal implementation, with `Schedule`/`ScheduleRecurring`/`Cancel` matching `TimerManager` so the manager name is the only thing that differs at a call site, and a separate manager rather than three more methods on `TimerManager` because calling the wrong one deadlocks. The Timer suite moves onto that route, and DevKernel gains `timers`, which fires one of each and reports both; nothing there exercised either timer API before. `SoftwareTimer` itself stays public and stays where it is: the member pass above already internalized everything on it that could corrupt the device registry, so what is left is a handle, and it is declared in `HAL.Interfaces` for a layering reason rather than a policy one, since `TimerDevice` ticks the registry from the interrupt path and `HAL.Interfaces` cannot reference the ring.

`HAL.Interfaces` ends at 25 entries: `IBlockDevice`, `MACAddress` and `SoftwareTimer` as a read-only handle.

A public member cannot expose another assembly's internal type even across `InternalsVisibleTo` (CS0050/CS0051), so the implementers follow: the two platform initializers and the two interrupt controllers go internal, and `Panic.CpuException`, whose callers all sit inside `Cosmos.Kernel`, goes with them. A public class may implement an internal interface from a friend assembly, verified with a throwaway build, so `X64PortIO`, `X64CpuOps`, `X64PowerOps` and the ARM64 counterparts could have stayed public; they are internal too, since nothing outside the HAL constructs one. Six grants make it compile: Core to `HAL.Interfaces` for the factory return types, `HAL.Interfaces` to HAL, the two arch HALs and System so the `LibraryInitializer` can still register what it enumerates, Core.X64 to HAL.X64 for the interrupt controller, which Core.ARM64 already had, and System to the Timer and Interrupts suites, which assert that a timer device got registered. `IRQContext`'s per-arch `.editorconfig` exclusion goes away with it, since an internal type never reaches the tracking files.

Dispatch resolve now throws the BCL `System.Runtime.AmbiguousImplementationException` instead of Core's own same-named copy, which is deleted. Standard code catching the BCL type never caught the kernel's throw before; anything catching `Cosmos.Kernel.Core.Runtime.AmbiguousImplementationException` stops compiling.

**The scheduler seam (`COSMOS0001`).** The policy interface, `SchedulerManager`, `Thread`, `PerCpuState`, `SchedulerExtensible`, `InterruptMaskScope` and the two enums become a real, marked seam, with `MaskInterrupts()` added so a policy can guard the entry points the manager does not. A seam with one implementation proves nothing, so it gets a second policy: `RoundRobinScheduler` lives in the Threading suite written the way a user kernel would write one, public surface only, no Core internals.

It found a real bug. `UpdateGlobalPass` divided a `Stopwatch` tick delta by a nanosecond quantum, so global virtual time ran several times fast on x64 and slow on ARM64, and the anti-starvation floor in `OnThreadYield` then snapped every runnable thread up to `GlobalPass`. Stride was not proportional at all — the four-times-tickets thread got an equal share. Fixed here, with a bound on the catch-up delta and the pre-sizing that keeps `InsertByPass` from allocating inside the tick.

**The packet seam (`COSMOS0002`).** The packet layer does not go internal. Crafting and reading raw packets is a real thing kernels do, so the protocol types (`EthernetPacket`, ARP, `IPPacket`, ICMP, `UdpPacket`, DHCP, DNS, `TcpPacket`), `NetworkStack.Send`/`HandlePacket` and the client members that take or return packets stay public under `[Experimental("COSMOS0002")]` — 223 tracked entries. Only the TCP machinery behind them (`Tcp`, `TransmissionControlBlock`, `Status`, `OutgoingBuffer`) went internal. The Network suite covers the seam from a consumer's position.

**`Canvas3D`.** `SvgaII3DCanvas` (renamed from `SVGAII3DCanvas` by the acronym pass) was the only symbol pulling concrete HAL types (`PciDevice`, `SvgaIIDriver`, `VMWareSVGAII3D`) into System's surface, so it went internal like the already-internal `GopCanvas` — and the 3D capability it was the only way to reach came back as a designed API instead. `Canvas3D`, `Camera3D`, `Mesh` and `Texture` are the public 3D surface (35 tracked entries, 38 counting the three `FullScreenCanvas` accessors), obtained from `FullScreenCanvas.GetFullScreenCanvas3D`, with the SVGA3D backend behind it. Its generic capabilities moved to the base at the same time: `Canvas.CopyPixels` is a new virtual with a clipped, overlap-safe software implementation that every canvas inherits (the SVGA override keeps the device's accelerated copy), and `Canvas.MovePixel` moves to the base with a software implementation, so both now work on every canvas instead of only VMware. `GetPixel` deleted (duplicate of the existing `GetPointColor` override).

**Build enforcement.** API tracking extends past System to `Core`, `HAL` and `HAL.Interfaces`, with `RS0016`/`RS0017` promoted to build errors, so any surface movement lands in the PR as a `PublicAPI.Unshipped.txt` diff. `RS0017` needed `WarningsAsErrors`: it has no source location, so the `.editorconfig` `[*.cs]` severity never applied and stale txt entries passed green builds — the hardened guard caught eight leftovers. The whole `CosmosTrackPublicApi` property block (this promotion plus `EnablePackageValidation`, inert since #447) had to move to `Directory.Build.targets`: `Directory.Build.props` is imported before the project body sets the flag, so property conditions on it never fired there. RS0016 stays editorconfig-driven (`/warnaserror` would trample the per-file exclusions). #452 merged in brings the 229 missing XML docs and CS1591 promoted to a build error on the tracked projects.

Package validation, running for real for the first time, failed the Release multi-arch pack with PKV0001 (no compile-time asset for net10.0). That layout is intentional: the package ships only `runtimes/linux-{x64,arm64}/lib/` assemblies and `build/*.props` selects the reference by RID (`Directory.MultiArch.targets`). Recorded in a `CompatibilitySuppressions.xml` next to each multi-arch project (System, Core and HAL) so the rest of the validation stays armed.

## Surface movement

Tracked entries, excluding each file's `#nullable enable` header, measured against the PR base:

| File | Before | After |
|---|---|---|
| `Cosmos.Kernel.System` | 1077 | 1052 |
| `Cosmos.Kernel.Core` | untracked | 81 |
| `Cosmos.Kernel.HAL` | untracked | 111 |
| `Cosmos.Kernel.HAL.Interfaces` | untracked | 25 |

System's net figure understates both directions: 458 entries left and 433 arrived. 125 of those pairs are the same member respelled by the acronym normalization or newly carrying the `[COSMOS0002]` seam prefix, so net of that wash 333 members genuinely left the surface and 308 arrived, the arrivals being the `Diagnostics` facades, `Canvas3D`, `KernelFeatures` and the field-to-property conversions. `PublicAPI.Shipped.txt` is still empty in all four; nothing is frozen yet.

Tracking still stops at those four assemblies, and docfx does not: it also generates API pages for `Cosmos.Kernel.HAL.X64`, `Cosmos.Kernel.HAL.ARM64`, `Cosmos.Kernel.Boot.Limine` and `Cosmos.Kernel.Debug`, so whatever stays public there is published as if it were kernel API. Review caught `PS2Ports` that way, the one public type the arch-HAL pass missed, with all 34 of its call sites inside its own assembly; it is internal now and both arch HALs are at zero. What is left public and untracked, for a follow-up: `Boot.Limine` 29 types (the boot protocol structs, which the test framework consumes), `Core.X64` 14 and `Core.ARM64` 8 (`LocalApic`, the GIC drivers, ACPI table readers, `DeviceMapper`), `Plugs` 20, `Cosmos.Kernel` 2 and `Debug` 1. None of them are in docfx except `Boot.Limine` and `Debug`.

## Verification

Policy behaviour, measured as the loop count of two CPU-bound spinners over a fixed window. One ratio separates the two scheduler policies from both sides — the four-times-tickets thread must clear 1.5x under Stride, and neither thread may reach it under Round-Robin:

| Policy | Tickets | x64 | ARM64 |
|---|---|---|---|
| Stride | 400 : 100 | 2.07 – 2.20 | 1.92 – 2.13 |
| Round-Robin | 100 : 100 | 0.88 – 1.13 | 1.00 – 1.06 |
| Round-Robin | 400 : 100 | 0.89 – 1.13 | 0.95 – 1.00 |

Suites run locally at this head, before the contract internalization, which changes no runtime behaviour and needs a rebuild to confirm green:

| Suite | Result | Covers |
|---|---|---|
| Threading | 73/73 | The scheduler seam, both policies, the Stride tick-unit fix |
| Network | 36/36 | Sockets/DHCP/DNS/ICMP through patched plugs referencing the internalized stack, plus the packet seam from a consumer's position |
| Graphic | 46/46 | `Canvas3D` discovery, the SVGA3D FIFO wire format, the `SvgaII3DCanvas` internalization and the `CopyPixels` lift |
| GarbageCollector | 45/45 | The TLAB path through the scheduler's `Thread` |
| Timer | 24/24 | The `TimerDevice` base through the new `InternalsVisibleTo` grant |
| Runtime | 103/103 | Dispatch resolve and the vendored metadata readers |
| Fat | 41/41 | White-box layout tests through `InternalsVisibleTo` |
| File | 37/37 | The VFS contracts and the disposable directory handle |
| Storage | 263/272, 9 skipped | The hoisted partition entry types |
| HelloWorld | 3/3 | Console through the internalized `ConsoleStream`/`ConsoleTextWriter` |

Beyond the suites: `make setup` is green end to end with package validation active, `make build` produces the DevKernel ISO with no `InternalsVisibleTo` grant (the ring-policy enforcement test), a scripted QEMU session drives `netconfig`/`netsend`/`netlisten`/Esc/`netsend` and both sends hit the wire, `grep` finds no public types left under the vendored directories, and no caller of a removed or internalized symbol survives outside its own assembly across src, tests, examples and docs.

The Threading cells split by what they can prove. The run-structure invariants — tail enqueue, head pick, quantum accounting, block, yield, exit — drive the hooks directly on a throwaway `PerCpuState`, so they need no timer and cannot flake. Only what genuinely needs a running kernel is measured live, and the live half deliberately does not assert dispatch order: a thread preempted inside its dispatch preamble goes to the tail, so the order threads reach their delegate is not the order they became ready.

## Review pass

A last pass over the branch found four things worth fixing before anything freezes, all in this PR's own new surface, plus the dead platform surface described above:

- `SchedulerInfo.RequestKill` only walked `CurrentThread` and the run queues, so killing a thread that `cpustat` had just listed as `Blocked` returned `NotFound`. It resolves against the thread registry now, and a blocked thread gets its own answer instead of being denied existence.
- `InterruptMaskScope` is a struct, so `default(InterruptMaskScope).Dispose()` restored flags nobody saved — clearing IF on x64, unmasking on ARM64, inside the caller's critical section. Only a scope handed out by `MaskInterrupts()` restores anything now.
- `PerCpuState.LastTickAt` and `Thread.CreatedAt` have no writer anywhere and never had one. Deleted rather than documented and frozen.
- `MACAddress.ToNumber`/`To32BitNumber` shifted `byte` values by 40 and 32, which C# masks to 8 and 0, so byte[0] landed on byte[4] and two adapters differing only in those bytes collided in `NetworkStack.MACMap`. `ToNumber` also sign-extended whenever byte[2] had its high bit set, returning `0xFFFFFFFFCCDDEEFF` for `AA:BB:CC:DD:EE:FF`. The `// TODO check shifting` under both is now answered.
- `MACAddress.Equals` threw `ArgumentException` for anything that was not a `MACAddress`, null included, where `object.Equals` requires `false`; `CompareTo(null)` threw too, where `IComparable` expects a positive result. `GetHashCode` built `GetType().AssemblyQualifiedName + "|" + ToString()` and hashed the string: reflection and two allocations to hash six bytes, in kernel code that is not allowed reflection. It folds the bytes now, consistently with `Equals`.

The scheduler plugging guide had the matching documentation bug: it recommended `GetSchedulerData<T>()`, which is a cast, when every user policy inherits the boot thread's `StrideThreadData` across the swap — following the checklist faulted inside the timer interrupt on the first tick. It says `as` now, and step 3 no longer points readers at an install path (`LibraryInitializer`) that is internal to the kernel.


src/Cosmos.Kernel.Core/Scheduler/AlarmSystem.cs
```diff
@@ -13,7 +13,7 @@ namespace Cosmos.Kernel.Core.Scheduler;
 /// primitives. Requires the scheduler; resolution is bounded by the scheduler
 /// tick.
 /// </summary>
-public static class AlarmSystem
+internal static class AlarmSystem
```
Guillermo-Santos, line 19: I think there isn't any issue with this class being public, @zarlo  what do you think?

valentinbreiz: I'm following the only Cosmos.Kernel.System API patern with a created Cosmos.Kernel.Sysyem.Timer.AlertManager that calls Core.
This is not decided it can be changed if we want to open more Core or HAL rings

zarlo: Other parts will likely want to have access to it later on, but for now we should default with internal. Also, for what this was initially added for, it's perfectly fine being internal.

src/Cosmos.Kernel.System/Graphics/Canvas.cs
```diff
@@ -517,10 +598,6 @@ public virtual void DrawLine(Color color, int x1, int y1, int x2, int y2)
     /// <param name="radius">The radius of the circle to draw.</param>
     public virtual void DrawCircle(Color color, int xCenter, int yCenter, int radius)
     {
-        ThrowIfCoordNotValid(xCenter + radius, yCenter);
-        ThrowIfCoordNotValid(xCenter - radius, yCenter);
-        ThrowIfCoordNotValid(xCenter, yCenter + radius);
-        ThrowIfCoordNotValid(xCenter, yCenter - radius);
```
valentinbreiz: This may should be kept for optimization instead of per point edge check

src/Cosmos.Kernel.System/Graphics/Canvas.cs
```diff
@@ -601,10 +678,6 @@ public virtual void DrawFilledCircle(Color color, int x0, int y0, int radius)
     /// <param name="yR">The Y radius.</param>
     public virtual void DrawEllipse(Color color, int xCenter, int yCenter, int xR, int yR)
     {
-        ThrowIfCoordNotValid(xCenter + xR, yCenter);
```
valentinbreiz: Same as https://github.com/valentinbreiz/nativeaot-patcher/pull/451/changes#r3884886359

src/Cosmos.Kernel.System/Graphics/GopCanvas.cs
```diff
@@ -272,15 +272,20 @@ public override void Clear(Color aColor)
...
+    // Every canvas clips at DrawPoint: a pixel outside the canvas is dropped,
+    // never drawn and never thrown for. Skipping the test is not the saving it
+    // looks like, because the same comparison then runs four times per pixel
+    // inside the driver's byte indexer, and a negative X wraps into the
+    // previous row instead of vanishing.
     public override void DrawPoint(Color aColor, int aX, int aY)
     {
         ThrowIfDriverNotInitialized();
 
+        if (aX < 0 || aX >= Width || aY < 0 || aY >= Height)
```
valentinbreiz: Same as https://github.com/valentinbreiz/nativeaot-patcher/pull/451/changes#r3884886359

src/Cosmos.Kernel.System/Graphics/GopCanvas.cs
```diff
@@ -324,7 +329,11 @@ public override void DrawPoint(Color aColor, int aX, int aY)
     public override void DrawPoint(uint aColor, int aX, int aY)
     {
         ThrowIfDriverNotInitialized();
-        //uint offset;
+
```
valentinbreiz: Same as https://github.com/valentinbreiz/nativeaot-patcher/pull/451/changes#r3884886359

PR #464, opened 2026-09-02
 Chart-based GC and CPU monitors in DevKernel

warquys:
I rewrote the monitors on my sandbox kernel. I pasted here because i like the visual and is a good exemple.
Moving the calls of drawligne in an helper class, help to understand who to create a graph (like in cpustat) and use drawligne.
I also move away the CPU stress from the cpu monitor and rewrote the cpu monitor to use the help method.
I added a graph for viewing the memory in the gcstat and added maker when the GC occurent.


warquys:
Fix order:
```cs
using System.Diagnostics;
using System.Drawing;
```
git amended and force with lease

zarlo (approved):

PR #472, opened 2026-09-16
Skip EfiRtc Initialize on x64 HAL

CodeCs1:
I've seen that new version still keep EfiTime on x64 HAL, so I'm creating PR on my own. It's a simple fix but it can now be able to run on UEFI system normally without getting `#PF` on boot (I've tested it).

Here is a thing, Limine STILL keep EFI Runtime Service (EfiRS) but some how all EfiRS functions  , including `GetTime()` function, assign to a physical memory address that lower than Limine HHDM address (correct me if i'm wrong)

I've also test it in a minimal C limine kernel (have mapped in Limine HHDM address) and still got same result after running `GetTime()` function.

Hope get it fixed.


Guillermo-Santos:
I think solution should another thing other than commenting this out.

CodeCs1:
> I think solution should another thing other than commenting this out.

There is no other way beside that. I could just create a Plug for it but I just too lazy for copy & patse plug for every new project.

Guillermo-Santos:
> > I think solution should another thing other than commenting this out.
> 
> There is no other way beside that. I could just create a Plug for it but I just too lazy for copy & patse plug for every new project.

A feature flag may be better.

CodeCs1:
> > > I think solution should another thing other than commenting this out.
> > 
> > There is no other way beside that. I could just create a Plug for it but I just too lazy for copy & patse plug for every new project.
> 
> A feature flag may be better.

This is more like a bug than a feature 🙂

valentinbreiz:
Thanks for the report, it reproduces under OVMF with the Timer suite ISO: `#PF` (error code 0x2, CR2 = 0x17) right after `[RTC] Initializing...`.

The physical addresses you saw are expected. The kernel declares no Limine base revision, so it runs at revision 0, which identity-maps every memory map region, and Limine never calls `SetVirtualAddressMap`. `GetTime` is reachable at its physical address.

The actual defect is the calling convention. EFIAPI on x86-64 is the Microsoft x64 convention (arguments in RCX/RDX above a 32-byte shadow area). The kernel is compiled for linux-x64, so `delegate* unmanaged` emitted a System V call with the `EFI_TIME*` in RDI and the null capabilities pointer in RSI; the firmware wrote its result through whatever RCX held. ARM64 never showed it because EFIAPI there is AAPCS64, the convention NativeAOT already uses.

The fix routes the call through a small System V to Microsoft x64 thunk (`_native_efi_call`, imported once in Core as `EfiNative.Call` like the other `_native_*` routines, with a pass-through of the same name on ARM64), and makes the `GetTime` field in `EfiRuntimeServices` an untyped pointer so a direct call cannot come back. With it the same ISO boots under OVMF, logs `[RTC] EFI GetTime status: 0` with the host UTC time, and passes 26/26; restoring the direct call brings the page fault back. No plug or feature flag needed on your side, the EFI clock now works on UEFI x64.

The fix is #473.

